PR #1, opened 2017-10-17
Init

fgimenez:

Kickoff of the project, only relevant the CI setup


CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
+# How to contribute
+
+shipyard is Apache 2.0 licensed and accepts contributions via GitHub pull requests. This document outlines some of the conventions on commit message formatting, contact points for developers and other resources to make getting your contribution into shipyard easier.
```
JosephSalisbury: not `shipyard`, right?

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+RUN mkdir -p /opt
+ADD ./e2e-harness /opt/e2e-harness
+
+RUN mkdir -p /opt/resources
+ADD resources/templates/ /opt/resources/templates
+
+RUN \
+    apk -Uuv add --update \
+      build-base \
```
JosephSalisbury: Can we specify particular versions here? Floating versions makes me nervous

README.md
```diff
@@ -0,0 +1,27 @@
+# e2e-harness
+
+[![CircleCI](https://circleci.com/gh/giantswarm/e2e-harness.svg?style=svg)](https://circleci.com/gh/giantswarm/e2e-harness)
```
JosephSalisbury: I know mr timothy prefers `style=shield`, if you wouldn't mind updating

cmd/root.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	RootCmd = &cobra.Command{
+		Use:   "e2e-harness",
+		Short: "Harness for custom kubernetes e2e testing",
+		Long:  ``,
```
JosephSalisbury: Is there a particular reason to not use " here?

JosephSalisbury: (this is more for my curiosity)

CONTRIBUTING.md
fgimenez: sure! fixed (there were a couple of them over there..)

Dockerfile
fgimenez: Indeed, very good catch, done

README.md
fgimenez: sure, done

cmd/root.go
fgimenez: Don't remember where did i got this from, i think that being the Long description backticks make sense if it's going to have more than one line, not the case :) using " now

fgimenez: Given that it is not being used I've removed the `Long` field

fgimenez:
@JosephSalisbury thx! comments addressed, ready for another pass

PR #2, opened 2017-10-19
Cluster setup teardown

fgimenez:

Basic setup of remote cluster and teardown. Also changes to `.circleci/config.yml`  to include a workflow, architect build as the first stage and then e2e stages, looks like this https://circleci.com/workflow-run/09a9000b-e91b-46d4-977a-7b3c3ac2d543


pkg/docker/docker.go
```diff
@@ -0,0 +1,35 @@
...
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/giantswarm/e2e-harness/pkg/harness"
+)
+
+func Run(imageTag string, out io.Writer, entrypoint string, args ...string) error {
+	args = append([]string{"quay.io/giantswarm/e2e-harness:" + imageTag}, args...)
```
kopiczko: Why not use shipyard as a library instead?

pkg/harness/harness.go
```diff
@@ -1,27 +1,80 @@
...
 
-type Harness struct {
-	logger micrologger.Logger
+// Status represents the current state of the
+// test harness. It is the token passed amongst
+// tasks.
+type Status struct {
+	RemoteCluster bool   `yaml:"remoteCluster"`
+	ImageTag      string `yaml:"imageTag"`
+}
```
kopiczko: > Status represents the current state

Looks like `State` is more appropriate name.
`ImageTag ` doesn't look like state. It looks like a configuration. Can we extend Task like:

```
type Task func(harness.Cofnig, *harness.Status) (*harness.Status, error)
```

pkg/tasks/tasks.go
```diff
@@ -0,0 +1,23 @@
+package tasks
+
+import "github.com/giantswarm/e2e-harness/pkg/harness"
+
+// Task represent a generic step in a pipeline
+type Task func(*harness.Status) (*harness.Status, error)
```
kopiczko: I don't see a point in passing pointer when the status is returned back.

pkg/harness/harness.go
```diff
@@ -1,27 +1,80 @@
...
+
+// Init initializes the harness
```
kopiczko: Since those functions are of `Task` type I'd like to see comments like `Init is a Task initializing the harness.`. This would make it easier to realise that. Also please try add full stops at the end <3

pkg/docker/docker.go
kopiczko: I understand you want to use the docker image as an aggregator of binaries. I think we can do better here.
My idea is to pass path to the binaries as CLI arguments defaulting to paths present in the image. That way we don't have to pass additional imageTag and make the code less confusing.

fgimenez:
> Why not use shipyard as a library instead?

The e2e-harness image is meant to provide a well-known working environment, something that we can test locally and have the same results as in CI. Having a fixed version of the shipyard binary in the image makes it possible to assess that it works and that all the required dependencies will be always in place (so that we don't need to rely on any package installed on the CI machine).

pkg/docker/docker.go
fgimenez: Sorry, I'm missing your point here. `quay.io/giantswarm/e2e-haeness` is built from this same repo, this is the Dockerfile in this PR https://github.com/giantswarm/e2e-harness/blob/946e47b607903afb304263c77eef281f77194b13/Dockerfile. It gives a common, well-known environment for all the repos using e2e-harness to run e2e-tests, atm it includes shipyard, kubectl and helm binaries, available in the default `PATH`. 

Now, the image tag is used to establish the version of the image being used, in the e2e-harness repo it will allow us to test the e2e-harness image just built by architect in the first CI step, in the clilent repos to fix the image version used to a well-known one. Why do we need to pass the path of the binaries? Why would this replace passing the imageTag?

pkg/harness/harness.go
fgimenez: Agree, `State` is much better. Btw, in the next PR there are types for each of the packages which are configurable, if you don't mind we can postpone this fix until then.

pkg/docker/docker.go
kopiczko: You download the binary for a version. Then you run it passing the same version as `image-tag` argument, because the binary runs other binaries from the image. That way you have to maintain two distribution channels, docker images, and binaries, which looks odd to me personally. Especially when I read the code. 

What I'm proposing is making this opposite. Instead of `image-tag` flag `e2e-harness` takes `shipyard-path`, `kubectl-path` and `helm-path` flags, defaulting to `/bin/shipyard`, `/bin/kubectl` and `/bin/helm` respectively. `e2e-harness` runs the binaries directly instead of invoking docker image of itself.  Then when you build the image you make sure binaries are in default paths,  you add `e2e-harness` as entrypoint and you can run `e2e-entrypoint` from a docker image. That way you avoid running docker inside the binary, you don't have two distribution channels, and the code looks more intuitive IMO.

If I'm still not clear we can sync on Monday and talk about this.

fgimenez: Ok, let's talk on monday, the problem I see with your approach is that every repo would need to know how to invoke the e2e-harness container, which paths to mount and where, which environment variables to pass, how to persist state between different calls, etc. Maybe I'm missing an easy way of doing it.

The current approach works similarly to architect, usually we would download `latest` release of the e2e-harness binary, as in https://github.com/giantswarm/aws-operator/blob/master/circle.yml#L11, and use a specific tag for the container image, as in https://github.com/giantswarm/architect/blob/master/workflow/helm.go#L16, anyway let's sync about this.

kopiczko: Now I understand. You also posted it here: https://github.com/giantswarm/e2e-harness/pull/3#discussion_r146103510

The reason you mentioned is good enough for me.

There is a little nice thing we can do about it though. We can get rid of `image-tag` argument by storing SHA during build time. You can create `gitCommit` variable in main.go and use that instead `image-tag` so we have that baked in into the binary. See https://github.com/giantswarm/architect/blob/f58f67262ac7828fe15350d51c118db3821a734c/workflow/golang.go#L239 for details.

WDYT?

It took me a while to wrap my head around your reasoning. Thanks for bearing with me <3

fgimenez: np at all! :) it has helped me a lot writing down all this things that were just floating around. `gitCommit` you mention sounds very nice,  /me looks...

pkg/harness/harness.go
fgimenez: Done

pkg/tasks/tasks.go
fgimenez: Done

pkg/docker/docker.go
fgimenez: Done the changes for using `gitCommit` and removing the flag, pls take look. As mentioned before things change a bit in later PRs (configurable values on each package, so that gitCommit can be passed to the docker value at creation time), now to keep things simple this var is injected into the `cmd` package from main.

main.go
```diff
@@ -7,7 +7,14 @@ import (
 	"github.com/giantswarm/e2e-harness/cmd"
 )
 
+var (
+	gitCommit = "latest"
```
kopiczko: We usually set it to `n/a`. https://github.com/giantswarm/kvm-operator/blob/328897482d9d15543dd4c50ea2449628a3cc316b/main.go#L23

PR #3, opened 2017-10-20
Declarative e2e test setup and teardown

fgimenez:
Requires #2 

With this changes it is possible to define declaratively what are the required steps to setup  and teardown a e2e test, they can be written in a `e2e/project.yaml` file like this:
```
version: 1
setup:
  - run: helm registry install quay.io/giantswarm/cert-operator-lab-chart -- --set imageTag=latest --set clusterName=my-cluster --wait
    waitFor:
      run: kubectl get certificate
      match: No resources found\.

  - run: helm registry install quay.io/giantswarm/cert-resource-lab-chart -- --set clusterName=my-cluster
    waitFor:
      run: kubectl get secrets
      match: my-cluster-api\s*Opaque
```
With this in place only the actual sonobuoy plugin execution is missing, will come next.


.circleci/config.yml
```diff
@@ -12,8 +12,50 @@ jobs:
 
     - run: ./architect build
 
+    - persist_to_workspace:
+        root: .
+        paths:
+        - ./e2e-harness
```
kopiczko: How e2e-harness is available in the CI? I don't see any download step. I think I'd be nice to run it as docker image.

fgimenez: It is build by architect in the first step. when used in another repo it will be downloaded, follows the same pattern as archiyect does. IMO using an image for this is a little bit cumbersomr, the client repos would need to know a lot of details about the inner workings, which directories to mount, which env vars to set, etc. The binary encapsulates these details, WDYT?

fgimenez:
@kopiczko review comments from the previous PR addressed, atm there's no state passed from one task to the next, all the values passed were, in fact, config and any of them was adding any change to the state, which makes things easier imo. Pls take a look.

fgimenez:
@kopiczko the declarative steps are only for setup and teardown, as stated in the PR title, I've found them very useful for preparing the aws-operator tests. In that case you need to install cert-operator, a given set of certificate custom objects and create an AWS VPC before installing the system under test, which is a chart. I think all this can be done with calls to CLI from go code, but find it much easier to define and understand in this declarative way.

The tests themselves are going to be written as sonobuoy plugins, as stated here https://github.com/giantswarm/giantswarm/issues/1853. In the next PR I'll upload an example of how this can be done, basically you prepare a Dockerfile which is used by the plugin, there you can put whatever you want, go code that reuses existing tests, k8s e2e conformance tests, whatever. You only need to take into account that this is going to run in-cluster and to write the results in a given format, using a package from e2e-harness will make this easy, WDYT?

fgimenez:
> I also don't think the tear down stage is important, as we don't care about the created container.

About teardown, this same project is meant to be used to execute e2e conformance tests on existing clusters, in that case we should clean up after finishing, let me know if that makes sense.

PR #4, opened 2017-10-24
Build and run tests

fgimenez:

in-cluster test in the example that runs in CI for this repo is disabled for now, I'm getting some errors when trying to retrieve the results from the sonobuoy worker, need to investigate. Other than that all is in place, now the pipeline looks like this: https://circleci.com/workflow-run/4b04e228-b837-4ba9-ad51-9ae7c65ffa84 in the `e2eTestBuild` stage the sonobuoy plugin is built, we can do this in parallel with the remote minikube setup. 

We have in place too the `e2eTestExecution` stage, where the in-cluster (sonobuoy) and out-of-cluster (declarative) tests are executed. Regarding this, currently there's no strict requirement about how test are written, we can write just interactions through kubectl, helm and other cli utilities, purely go (/python/bash/...) code in a sonobuoy plugin, or a combination of the two.


fgimenez:
Thx @r7vme! :) already fixed, this is how the sonobuoy plugin execution looks like https://circleci.com/gh/giantswarm/e2e-harness/152

PR #7, opened 2017-11-06
add init command

fgimenez:
Called from a project's root the init command creates a basic `e2e` directory to begin writing and running tests.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -1,0 +1,40 @@
+package cmd
+
+import (
+	"github.com/giantswarm/e2e-harness/pkg/harness"
+	"github.com/giantswarm/e2e-harness/pkg/initializer"
+	"github.com/giantswarm/e2e-harness/pkg/tasks"
+	"github.com/giantswarm/micrologger"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+)
+
+var (
+	InitCmd = &cobra.Command{
+		Use:   "init",
+		Short: "initialize project to develop and run k8s e2e tests",
+		RunE:  runInit,
+	}
+)
+
+func init() {
+	RootCmd.AddCommand(InitCmd)
+}
+
+func runInit(cmd *cobra.Command, args []string) error {
+	logger, err := micrologger.New(micrologger.DefaultConfig())
+	if err != nil {
+		return err
+	}
+
+	projectName := harness.GetProjectName()
+	fs := afero.NewOsFs()
+	i := initializer.New(logger, fs, projectName)
+
+	// tasks to run
+	bundle := []tasks.Task{
+		i.CreateLayout,
+	}
+
+	return tasks.Run(bundle)
+}
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -31,7 +31,7 @@ func runInit(cmd *cobra.Command, args []string) error {
 	fs := afero.NewOsFs()
 	i := initializer.New(logger, fs, projectName)
 
-	// tasks to run
+	// tasks to run.
 	bundle := []tasks.Task{
 		i.CreateLayout,
 	}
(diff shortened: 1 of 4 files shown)


cmd/init.go
```diff
@@ -0,0 +1,40 @@
...
+	logger, err := micrologger.New(micrologger.DefaultConfig())
+	if err != nil {
+		return err
+	}
+
+	projectName := harness.GetProjectName()
+	fs := afero.NewOsFs()
+	i := initializer.New(logger, fs, projectName)
+
+	// tasks to run
```
MarcelMue: Going to nitpic here:
I think that comments should always end with a full stop.

pkg/initializer/initializer.go
```diff
@@ -0,0 +1,236 @@
...
+)
+
+const (
+	DockerFileFmt = `FROM alpine:3.6
+
+ADD ./%s-e2e /e2e
+
+ENTRYPOINT ["/e2e"]
+`
+	MainGoContentFmt = `package main
```
MarcelMue: I really dislike having go content in a constant here. I would much prefer to have it in an actual file which is included. This has to be possible. The current state just removes all our editor goodness from the code.

pkg/initializer/initializer.go
```diff
@@ -0,0 +1,236 @@
...
+func New(logger micrologger.Logger, fs afero.Fs, projectName string) *Initializer {
+	return &Initializer{
+		logger:      logger,
+		fs:          fs,
+		projectName: projectName,
+	}
+}
+
+func (i *Initializer) CreateLayout() error {
+	wd, err := os.Getwd()
```
MarcelMue: Why are you not using `afero` here? I think we use it everywhere else. Just curious.

pkg/initializer/initializer.go
fgimenez: Yep agree, but don't know how to do it differently, the only thing that comes to mind is code generation, but imo it's too complex just for dropping these snippets. It might involve using templates for the files too, so that they wouldn't be proper go files either.

However we are looking at them from the producer project, once you initialize the project from the consumer project point of view you will get the go files ready to use, let me know wdyt.

pkg/initializer/initializer.go
fgimenez: Yep, an afero.Fs member is in place in the `i` value and is used for the rest of file system operations, but Afero doesn't seem to have a `Getwd` file system method.

fgimenez:
@MarcelMue thx for your review! About the tests, I'm using subtests here https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks, I find them easier to group tests and define common setup and teardown actions for each of them without repeating.

pkg/initializer/initializer.go
MarcelMue: Mhm okay, I am not really into it that much but template files would be preferable to me. We can iterate on this though.

pkg/initializer/initializer.go
MarcelMue: Yeah you're right, maybe there is just no issue with `os.Getwd()` here, the more you know I guess :D

fgimenez:
> How is it related to local execution?

@kopiczko not too much in fact :) you are totally right, `init` creates a basic structure and you can use it for local or remote execution. That's why i created a separate PR, I wrote it with local execution because while writing the README it was pretty clear that this piece was missing.

pkg/initializer/initializer.go
xh3b4sd: The odd thing about separate template files in go is the way of building them into the compiled binary. There was a lot of hustle already where we tried different approaches. The less evil we found was having them in go code as strings.

MarcelMue: Yeah you're right. I don't like it but it seems like strings are the path of least resistance for now.

PR #9, opened 2017-11-13
execute on local minikube

fgimenez:
These changes will allow execution of tests using a local minikube. It should be started with RBAC enabled before setting up the e2e environment:
```
$ minikube start --extra-config=apiserver.Authorization.Mode=RBAC
```
Then, the common workflow would be:
```
$ e2e-harness init # only if ./e2e doesn't exists
[ hack on ./e2e directory... ]
$ e2e-harness setup --remote=false
$ e2e-harness test # repeat as needed
$ e2e-harness teardown
```

@kopiczko there's another branch in progress for running the out of cluster tests connecting directly to the test cluster, defined as golang tests that receive a configured k8s client but removing sonobuoy (used for in-cluster) and the project.yaml step definitions from the equation as you proposed. The same tests can be used for both local development and CI, will come next.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,11 @@ for each operation mode:
 
 * Local: [minikube](https://github.com/kubernetes/minikube) should be started with
 RBAC enabled before running e2e-harness:
+
 ```
 $ minikube start --extra-config=apiserver.Authorization.Mode=RBAC
 ```
+
 * Remote: as stated above e2e-harness uses [shipyard](https://giathub.com/giantswarm/shipyard)
 for setting up the remote cluster. shipyard currently only supports AWS as the
 backend engine, so the common environment variables for granting access to AWS
@@ -77,12 +79,18 @@ see below. This is how the e2e directory looks like this:
     ├── example.go
     └── runner.go
 ```
+
 `project.yaml` defines how the end to end tests should be setup, which tests to
 run and how to tear them down. It includes the following fields:
+* `setup`: contains an array of steps to set the testing environment up.
+* `outOfClusterTest`: set of steps to execute checks as an external client
+* `inClusterTest`: configuration for the tests to be run as sonobuoy plugins, see
+below.
+* `teardown`: steps required to clean up the test environment.
 
 The rest of the files are an example of in cluster tests. These are executed as
 sonobuoy plugins. The example is written in go but you could use any langauge,
-see [the sonobuoy examples section](). These are the files included:
+see [the sonobuoy examples section](https://github.com/heptio/sonobuoy/tree/master/examples). These are the files included:
 
 * `Dockerfile`: Required, this is the image definition for the sonobuoy plugin that
 runs the in-cluster tests, see below for details.
@@ -96,20 +104,25 @@ func init() {
   Add(TestExample)
 }
 ```
+
 The test functions need to fulfill this interface:
+
 ```
 type Test func() (description string, err error)
 ```
-being the `description` used in the sonobuoy test output.
+
+The `description` used in the sonobuoy test output.
 
 ## e2e-harness lifecycle
 
 An e2e test execution involves three stages:
 
 * Setup: this is performed with the `setup` command.
+
 ```
 $ e2e-harness setup --remote=[false|true]
 ```
+
 It takes care of:
   - Initializing the project, creating an interchange directory that will
   be mounted on the test container for keeping state between command
@@ -118,7 +131,7 @@ It takes care of:
   will involve the creation of the cluster too. In local executions, the
   connection settings to be able running minikube are made available in
   the test container interchange directory.
-  - Run common set up steps: these will pu the test cluster in a common
+  - Run common set up steps: these will put the test cluster in a common
   initial state, basically installing tiller (helm's server side part) including
   the required resources to make it work with RBAC enabled.
   - Run specific setup steps: this are defined in the project file
@@ -126,18 +139,22 @@ It takes care of:
   description above) and can do things like installing the chart under test,
   setting up required external resources, etc.
 * Run tests: invoking the `test` command:
+
 ```
 $ e2e-harness test
 ```
+
 First, the out of cluster tests are executed, if any, defined as
 regular `step` entries under the `outOfClusterTests` key. Then the in cluster tests,
 if they are enabled in `project.yaml`. You should enabled them only if they are
 present, if not enabling them causes an error, see above for a descrition about
 how to write this kind of tests.
 * Teardown: the teardown phase is executing using the `teardown` command.
+
 ```
 $ e2e-harness teardown
 ```
+
 It includes:
   - Run specific teardown steps: this are defined in the project file
   `e2e/project.yaml` under the `teardown` key. They are common steps (see their
@@ -158,11 +175,12 @@ Currently the out of cluster tests are executed as `step` elements in the
 `e2e/project.yaml` file, the elements in a `step` are:
 
 * `run`: string with the command to execute, multiline allowed.
-* `WaitFor`: optional element to check if the `run` command was successful, it has
+* `waitFor`: optional element to check if the `run` command was successful, it has
 as nested items a `run` entry with the check command to be executed, a `pattern`
 entry with a regular expression pattern to be checked against the previous command
 output and a `timeout` with the number of miliseconds to keep executing the command
 and checking the output. An example step could be:
+
 ```
 - run: kubectl create namespace giantswarm
   waitFor:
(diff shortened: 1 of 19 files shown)


pkg/cluster/cluster.go
```diff
@@ -21,9 +27,27 @@ func New(logger micrologger.Logger, runner runner.Runner, remoteCluster bool) *C
...
+		return err
+	}
+
+	err = c.copyMinikubeAssets(usr.HomeDir)
+	if err != nil {
+		return err
+	}
+	err = c.setupMinikubeConfig(usr.HomeDir)
+	if err != nil {
+		return err
```
xh3b4sd: I would like to see error masking everywhere.

pkg/cluster/cluster.go
```diff
@@ -39,3 +63,115 @@ func (c *Cluster) clusterAction(command string) error {
...
+	// to access minikube using $HOME/.minikube, we store this in originDir
+	originDir := filepath.Join(homeDir, ".minikube")
+
+	// path is the actual location of the k8s config file that will be used from the
+	// test container
+	path, err := getMinikubeConfigPath()
+	if err != nil {
+		return err
+	}
+	read, err := ioutil.ReadFile(path)
```
xh3b4sd: IMO we should abstract the FS magic using something like `afero`.

fgimenez: Indeed, it is used in other places, i'll update it to use everywhere.

pkg/cluster/cluster.go
fgimenez: Agree, i'll start by masking every error once in the closest place to where it originated, we can extend from there if needed.

README.md
```diff
@@ -10,7 +10,197 @@ Clone the git repository: https://github.com/giantswarm/e2e-harness.git
...
+```
+├── Dockerfile
+├── main.go
+├── project.yaml
+└── tests
+    ├── example.go
+    └── runner.go
+```
+`project.yaml` defines how the end to end tests should be setup, which tests to
+run and how to tear them down. It includes the following fields:
```
kopiczko: Fields description missing? also new line after ` ``` `

pkg/cluster/cluster.go
xh3b4sd: IMO only masking in the originating place is just not good enough. You will refactor code and forget to mask where the origin changes. You will also not see where the error is going through and when there are multiple options you will have a harder time to find out what is actually going on. As I said already. Feel free to suggest alternatives to the state of the art but as long as there is no alternative that meets all the requirements of debuggability I would like to stick with the current conventions. 

It was just on Saturday that I was on call and had to debug gsctl, which does not mask errors. It took me 30 minutes during an incident situation to find out what the fucking problem was and where the error came from. Considering that I am always asking to mask errors everywhere you can imagine that I was pretty pissed on the weekend when I had to dig into the depths of some project I am not used to.

README.md
```diff
@@ -10,7 +10,197 @@ Clone the git repository: https://github.com/giantswarm/e2e-harness.git
...
+
+The rest of the files are an example of in cluster tests. These are executed as
+sonobuoy plugins. The example is written in go but you could use any langauge,
+see [the sonobuoy examples section](). These are the files included:
```
kopiczko: I think the link is missing.

README.md
```diff
@@ -10,7 +10,197 @@ Clone the git repository: https://github.com/giantswarm/e2e-harness.git
...
+```
+func init() {
+  Add(TestExample)
+}
+```
+The test functions need to fulfill this interface:
+```
+type Test func() (description string, err error)
+```
+being the `description` used in the sonobuoy test output.
```
kopiczko: ```The `description` is used in the sonobuoy test output.```

README.md
```diff
@@ -10,7 +10,197 @@ Clone the git repository: https://github.com/giantswarm/e2e-harness.git
...
+```
+It takes care of:
+  - Initializing the project, creating an interchange directory that will
+  be mounted on the test container for keeping state between command
+  executions.
+  - Prepare the connection to the test cluster: for remote executions this
+  will involve the creation of the cluster too. In local executions, the
+  connection settings to be able running minikube are made available in
+  the test container interchange directory.
+  - Run common set up steps: these will pu the test cluster in a common
```
kopiczko: `pu` -> `put`

README.md
```diff
@@ -10,7 +10,197 @@ Clone the git repository: https://github.com/giantswarm/e2e-harness.git
...
+e2e tests can be executed from agents either out of the test cluster or running
+inside that cluster:
+
+### Out of cluster tests
+
+Currently the out of cluster tests are executed as `step` elements in the
+`e2e/project.yaml` file, the elements in a `step` are:
+
+* `run`: string with the command to execute, multiline allowed.
+* `WaitFor`: optional element to check if the `run` command was successful, it has
```
kopiczko: ``` `waitFor` ```

pkg/cluster/cluster.go
```diff
@@ -39,3 +63,115 @@ func (c *Cluster) clusterAction(command string) error {
...
+	out, err := os.Create(dst)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		cerr := out.Close()
+		if err == nil {
+			err = cerr
+		}
+	}()
```
kopiczko: What is this function supposed to do?

pkg/project/project.go
```diff
@@ -114,8 +113,36 @@ func (p *Project) CommonSetupSteps() error {
...
+			Run: "helm reset --force",
+		},
+		Step{
+			Run: "kubectl delete clusterrolebinding permissive-binding",
+		},
+		Step{
+			Run: "kubectl -n kube-system delete sa tiller",
+		},
+		Step{
+			Run: "kubectl delete clusterrolebinding tiller",
```
kopiczko: I think it'd be nice to use `--context=minikube` explicitly. I'm pretty sure many ppl will forget to set default context to minikube before running those tests.

pkg/cluster/cluster.go
fgimenez: > You will also not see where the error is going through and when there are multiple options you will have a harder time to find out what is actually going on

I agree with the convention but could you pls explain this? I feel I'm missing something, you get an error and it has attached the stack trace of where it was originated. All the calls that provoked that error are recorded there with file names and line numbers, what is the advantage of masking the error again as it bubbles up and the call stack unrolls?

> It was just on Saturday that I was on call and had to debug gsctl, which does not mask errors

Indeed, that's a bad situation, not masking at all should not happen.

fgimenez:
Thanks a lot for your feedback @kopiczko i'll address your comments shortly.

> I don't think they need to be run from the e2e-harness at all.

I assume you mean exected from the harness container. One of the main goals of all this project is having reliable test executions so that we can make sure that the test we run have the same results no matter where they are executed. Because of the constraints of the CI environment, e2e-harness aims to provide a stable, immutable environment from which the tests are executed, so that that goal can be met. If we run the same tests locally in an environment different from CI's the results are not so valuable.

Speaking of complexity, i think that having a single code path is simpler than having one for local execution and other different for CI executions. Not only for the code itself considered statically but also for its dynamics, imo that way it will be easier to maintain, debug, change, etc. The setup phase got more complex now because it must take into account the local execution, but the run phase is as simple as it was (it didn't change), and it should be the important phase for us, the results come from there. And it only got a little bit more complex imo, just to make minikube accessible from the test container. From the end user point of view, you just need to execute the same command with an additional flag.

Also, please bear with me, the last changes for local execution, which address some of your suggestions, are comming next, we can discuss further in that PR.

pkg/cluster/cluster.go
fgimenez: Done

README.md
fgimenez: Sure, added.

README.md
fgimenez: Done.

README.md
fgimenez: Yep, much better

README.md
fgimenez: Done

README.md
fgimenez: Done

pkg/cluster/cluster.go
fgimenez: Makes it possible to not swallow a possible error from the deferred `*File.Close` call

pkg/project/project.go
fgimenez: That wouldn't work with the helm commands, instead I've added a `kubectl config use-context minikube` as the first step in the common setup tasks, pls take a look

fgimenez:
@kopiczko thx!

PR #10, opened 2017-11-20
Remove in cluster tests

fgimenez:
These changes address part of the excellent feedback given so far, with them the sonobuoy tests are removed (a separate conformance project is on its way), only out of cluster tests are taken into account. I think it makes sense to not mix the conformance test execution with potentially shorter CI cycles. This by itself simplifies the test execution greatly, there's no need to build a Docker image for the tests, they can be regular go tests guarded by a build tag and, because of that, the dependencies of the e2e test don't affect unit tests execution times. 

The `e2e/project.yaml` is also simpler now, only configuration options for setting environment variables for the execution (with interpolation from the current environment available). 

What is kept from the previous implementation is the execution of the tests in a container, imo this is essential for having consistent results in different execution environments.

As before, after starting minikube with:
```
$ minikube start --extra-config=apiserver.Authorization.Mode=RBAC
```
the lifecycle will look like:
```
$ e2e-harness init # only if ./e2e doesn't exists
[ hack on ./e2e directory... ]
$ e2e-harness setup --remote=false
$ e2e-harness test # repeat as needed
$ e2e-harness teardown
```


MarcelMue:
I like this change, thanks for the explanation. Are we going to do sonobuoy conformance elsewhere now?

fgimenez:
@MarcelMue sure, I was planning to move all the code in this project that allowed us to configure, install, run and retrieve results from sonobuoy to a different repo, not yet in place though.

README.md
```diff
@@ -138,90 +85,71 @@ It takes care of:
...
 
-First, the out of cluster tests are executed, if any, defined as
-regular `step` entries under the `outOfClusterTests` key. Then the in cluster tests,
-if they are enabled in `project.yaml`. You should enabled them only if they are
-present, if not enabling them causes an error, see above for a descrition about
-how to write this kind of tests.
+First, the binary and docker image for the project are built. Then the tests
+are compiled, not executed directly. Finally, that binary is executed in the
+test container with an environment defined by the environment variables set
+in the `project.yam` file.
```
MarcelMue: Typo: `project.yaml`

README.md
```diff
@@ -138,90 +85,71 @@ It takes care of:
...
 
-It includes:
-  - Run specific teardown steps: this are defined in the project file
-  `e2e/project.yaml` under the `teardown` key. They are common steps (see their
-  description above) and can do things like removing the chart under test,
-  tearing down required external resources, etc.
-  - Run commomn tear down steps: these differ depending on the mode of operation,
-  for remote, ephemeral clusters they are just deleted, for local clusters tiller
-  and all the required RBAC setup is removed.
+It consists of run commomn tear down steps: these differ depending on the mode
```
MarcelMue: Better english "It consists of commonly run tear down steps" - I am not exactly sure what you mean here.

README.md
```diff
@@ -138,90 +85,71 @@ It takes care of:
...
 
-### In cluster tests
+## Writing tests
 
-The in cluster tests execution are controlled by the `inClusterTests` entry in the project file:
+e2e tests are executed from the test container and are regular go test, for writing them
+just keep in mind these considerations:
 
+* The tests will be executed from the test container, this is it's [Dockerfile](https://github.com/giantswarm/e2e-harness/blob/master/Dockerfile).
+* All the go files should be guarded by a `e2e` build tag, being their first line:
```
MarcelMue: Just for my understanding: This means, that the very first line in my test file should be `// build + e2e` right?

e2e/client.go
```diff
@@ -0,0 +1,24 @@
+// +build e2e
```
MarcelMue: Here you start with `+` in the readme it's `build + e2e`

README.md
fgimenez: Yep, the first line should be `// +build e2e` (wrong here, right below)

e2e/client.go
fgimenez: Yes, this one is ok (== verified on ci :)

README.md
fgimenez: Done

README.md
fgimenez: Yep, finally `consists of running common steps...`

e2e/client.go
kopiczko: FYI it's k8srequired now.

PR #14, opened 2018-02-01
Setup localkube

fgimenez:
Allows to setup localkube (minikube with `--vm-driver=none`) on circleCI machines. This way we don't need to spin up a remote minikube for each e2e execution and the setup is slightly quicker (~30sec see https://circleci.com/workflow-run/7c72def2-b09f-49bc-bc30-f07a5334ab46 for example).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,27 +24,30 @@ jobs:
             ./architect deploy
           fi
 
-  e2eSetup:
+  e2eLocalTest:
+    working_directory: /home/circleci/.go_workspace/src/github.com/giantswarm/e2e-harness
     machine: true
+    environment:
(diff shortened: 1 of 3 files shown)


kopiczko:
This `Allows to setup localkube`. Does that mean we don't need remote clusters at all?

fgimenez:
It depends on how setting up all the requirements for the concrete e2e test work on circleCI machines, I'm concerned mostly about memory. Initial tests with cert-operator worked just fine https://circleci.com/gh/giantswarm/cert-operator/656, there the test machine was running localkube with pods for vault and cert-operator. Let's see how it goes after adding to the equation aws-operator, if we have OOM there we can keep using the remote cluster for that concrete case.

kopiczko:
I'm full of hope ;)

PR #18, opened 2018-02-09
Cleanup CircleCI config

indyfree:
Some minor cleanup 
- Fixed indentation
- Changed job naming to circleci convention (as seen in the docs)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,26 +3,26 @@ jobs:
   build:
     machine: true
     steps:
-    - checkout
+      - checkout
 
-    - run: |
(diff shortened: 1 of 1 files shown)


indyfree:
Thanks for the review! 
I agree with the indentation issue, especially if its already consistent across all the other repos, lets stick to this. 
Regarding naming, dashes would be nice, but is not critical by any means, plus camel-case is used in the other repos too, I would be fine with it. I am thinking of revoking this PR, WDYT?

fgimenez:
Up to you, naming with dashes is ok but requires changes, if you think they are worth the effort go ahead! :)

indyfree:
Think it's fine this way. No need to tinker around in other repos

PR #19, opened 2018-02-21
moved to dep

fgimenez:


Gopkg.toml
```diff
@@ -0,0 +1,58 @@
...
+  name = "k8s.io/apimachinery"
+  version = "kubernetes-1.8.5"
+
+[[constraint]]
+  name = "k8s.io/client-go"
+  version = "kubernetes-1.8.5"
+
+[prune]
+  go-tests = true
+  unused-packages = true
```
kopiczko: I didn't know that section! I see ton of stuff removed.

fgimenez: yep :) just `dep init` with latest version did it.

PR #21, opened 2018-02-22
Add framework package

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2440 and https://github.com/giantswarm/giantswarm/issues/2441

Extracted code from aws-operator e2e tests, this can be reused in other projects.


Gopkg.toml
```diff
@@ -41,6 +45,11 @@
   branch = "master"
   name = "github.com/spf13/cobra"
 
+[[constraint]]
+  branch = "master"
+  name = "github.com/cenkalti/backoff"
+
+
```
xh3b4sd: I would like to sort this alphabetically. There is also an additional empty line.

pkg/framework/error.go
```diff
@@ -0,0 +1,31 @@
+package framework
+
+import "github.com/giantswarm/microerror"
+
+var waitTimeoutError = microerror.New("waitTimeout")
```
xh3b4sd: Same thing with the order here.

Gopkg.toml
fgimenez: Sure done.

pkg/framework/error.go
fgimenez: Done.

PR #22, opened 2018-03-05
Only build binaries if they are not already created

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2712 and https://github.com/giantswarm/giantswarm/issues/2678

With these changes we will be able to reuse the binary created in the initial build job, improving the e2e total build time.

### DIFF
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -59,6 +59,13 @@ func (c *Compiler) CompileTests() error {
 }
 
 func (c *Compiler) compileMain(binaryName, path string) error {
+	// do not build if binary is already there
+	binPath := filepath.Join(path, binaryName)
+	if executebleExists(binPath) {
+		c.logger.Log("info", "main binary exists, not building")
+		return nil
+	}
+
 	cmd := exec.Command("go", "build", "-o", binaryName, ".")
 	cmd.Env = append(os.Environ(), "CGO_ENABLED=0", "GOOS=linux")
 	cmd.Dir = path
@@ -67,6 +74,13 @@ func (c *Compiler) compileMain(binaryName, path string) error {
 }
 
 func (c *Compiler) compileTests(binaryName, path string) error {
+	// do not build if binary is already there
+	binPath := filepath.Join(path, binaryName)
+	if executebleExists(binPath) {
+		c.logger.Log("info", "test binary exists, not building")
+		return nil
+	}
+
 	cmd := exec.Command("go", "test", "-c", "-o", binaryName, "-tags", "k8srequired", ".")
 	cmd.Env = append(os.Environ(), "CGO_ENABLED=0", "GOOS=linux")
 	cmd.Dir = path
@@ -74,4 +88,19 @@ func (c *Compiler) compileTests(binaryName, path string) error {
 	cmd.Stdout = os.Stdout
 
 	return cmd.Run()
+}
+
+func executebleExists(path string) bool {
+	fi, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+
+	if fi.IsDir() {
+		return false
+	}
+
+	isExecutable := fi.Mode()&os.FileMode(0111) != 0
+
+	return isExecutable
 }
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -108,8 +108,8 @@ func executebleExists(path string) bool {
 	// 0111 octal represents a mode with the executable bit set
 	// for user, group and others. Performing a bitwise and with
 	// the mode of the file would only result in 0 if all these
-	// bit flags are 0, so if the result of the and is different
-	// from 0 we can assume that the file is executable.
+	// bit flags are 0, so if the result is different from 0 we
+	// can assume that the file is executable.
 	isExecutable := fi.Mode()&os.FileMode(0111) != 0
 
 	return isExecutable
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -58,6 +58,8 @@ func (c *Compiler) CompileTests() error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/compiler/compiler.go
```diff
@@ -59,6 +59,13 @@ func (c *Compiler) CompileTests() error {
 }
 
 func (c *Compiler) compileMain(binaryName, path string) error {
+	// do not build if binary is already there
```
MarcelMue: Can you, instead of having a comment here, have some godoc which explains what happens here in one or two sentences?

pkg/compiler/compiler.go
```diff
@@ -67,6 +74,13 @@ func (c *Compiler) compileMain(binaryName, path string) error {
 }
 
 func (c *Compiler) compileTests(binaryName, path string) error {
+	// do not build if binary is already there
+	binPath := filepath.Join(path, binaryName)
+	if executebleExists(binPath) {
+		c.logger.Log("info", "test binary exists, not building")
+		return nil
```
MarcelMue: Does `Run()` behave well with `nil` in the bundle to execute? I would be concerned here.

pkg/compiler/compiler.go
```diff
@@ -59,6 +59,13 @@ func (c *Compiler) CompileTests() error {
 }
 
 func (c *Compiler) compileMain(binaryName, path string) error {
+	// do not build if binary is already there
+	binPath := filepath.Join(path, binaryName)
+	if executebleExists(binPath) {
+		c.logger.Log("info", "main binary exists, not building")
```
xh3b4sd: As we discussed in the last lunch talk, there is a new logging format which we should use in order to leverage more nifty EFK stack features.

pkg/compiler/compiler.go
```diff
@@ -75,3 +89,18 @@ func (c *Compiler) compileTests(binaryName, path string) error {
...
+	fi, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+
+	if fi.IsDir() {
+		return false
+	}
+
+	isExecutable := fi.Mode()&os.FileMode(0111) != 0
```
xh3b4sd: That looks super magic to me. Some information on what is going on here would be nice.

fgimenez: Sure, added some comments, PTAL.

pkg/compiler/compiler.go
fgimenez: Great, where can I read the details?

pkg/compiler/compiler.go
fgimenez: Indeed, that's the expected behavior, a `Task` returns `error` https://github.com/giantswarm/e2e-harness/blob/master/pkg/tasks/tasks.go#L6 and if the error returned is nil `Run` keeps processing tasks https://github.com/giantswarm/e2e-harness/blob/master/pkg/tasks/tasks.go#L12.

pkg/compiler/compiler.go
xh3b4sd: We still need to write it down. There have been PRs though in which we introduced this. The easiest out there is something like his: https://github.com/giantswarm/operatorkit/blob/39f1a4387b0e8ffb148f54ca8b15eee7af537ca6/framework/framework.go#L555.

pkg/compiler/compiler.go
```diff
@@ -75,3 +89,23 @@ func (c *Compiler) compileTests(binaryName, path string) error {
...
+	}
+
+	if fi.IsDir() {
+		return false
+	}
+
+	// 0111 octal represents a mode with the executable bit set
+	// for user, group and others. Performing a bitwise and with
+	// the mode of the file would only result in 0 if all these
+	// bit flags are 0, so if the result of the and is different
```
xh3b4sd: `result of the and is` -> `result is`

pkg/compiler/compiler.go
```diff
@@ -75,3 +89,23 @@ func (c *Compiler) compileTests(binaryName, path string) error {
...
+	// from 0 wee can assume that the file is executable.
```
xh3b4sd: `wee` -> `we`

pkg/compiler/compiler.go
fgimenez: Done, PTAL.

pkg/compiler/compiler.go
fgimenez: Done

pkg/compiler/compiler.go
fgimenez: Done

pkg/compiler/compiler.go
kopiczko: `err == nil` is always good :D

pkg/compiler/compiler.go
```diff
@@ -75,3 +94,23 @@ func (c *Compiler) compileTests(binaryName, path string) error {
...
+	if fi.IsDir() {
+		return false
+	}
+
+	// 0111 octal represents a mode with the executable bit set
+	// for user, group and others. Performing a bitwise and with
+	// the mode of the file would only result in 0 if all these
+	// bit flags are 0, so if the result is different from 0 we
+	// can assume that the file is executable.
+	isExecutable := fi.Mode()&os.FileMode(0111) != 0
```
kopiczko: This is my favorite line :)
Including description.

pkg/compiler/compiler.go
MarcelMue: Yes, I messed up here - probably shouldn't be doing reviews so late in the evening :D

pkg/compiler/compiler.go
fgimenez: Nice, done PTAL.

indyfree:
Already some numbers how much time that saved us? :relaxed:

fgimenez:
Not a full aws-operator clean build yet, let's see :smile:

PR #26, opened 2018-03-09
adds --test-dir flag to configure sub folders

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2774.


cmd/setup.go
```diff
@@ -20,6 +20,9 @@ var (
 		Short: "setup e2e tests",
 		RunE:  runSetup,
 	}
+)
+
+var (
```
xh3b4sd: I think it makes sense to separate the command and its flags in different blocks.

cmd/test.go
```diff
@@ -23,8 +23,14 @@ var (
...
 )
 
+var (
+	testDir string
+)
+
 func init() {
 	RootCmd.AddCommand(TestCmd)
+
+	TestCmd.Flags().StringVar(&testDir, "test-dir", "integration", "Name of the directory containing executable tests.")
```
xh3b4sd: Adding a test dir flag which defaults to the currently used value. That way we can structure different tests in different directories and run them in parallel OOB by plain circle support.

cmd/test.go
```diff
@@ -65,7 +71,16 @@ func runTest(cmd *cobra.Command, args []string) error {
 	}
 	p := project.New(pDeps, pCfg)
 
-	comp := compiler.New(logger)
+	var comp *compiler.Compiler
```
xh3b4sd: Applying our usual style here. I scan scan code faster when it is consistent and this is also useful since we now have multiple inputs for the constructor.

pkg/compiler/compiler.go
```diff
@@ -46,15 +58,16 @@ func (c *Compiler) CompileTests() error {
 		return microerror.Mask(err)
 	}
 
-	name := harness.GetProjectName()
+	e2eBinary := harness.GetProjectName() + "-e2e"
+	e2eDir := filepath.Join(dir, c.testDir)
 
-	e2eBinary := name + "-e2e"
-	e2eDir := filepath.Join(dir, "integration")
```
xh3b4sd: `integration` is the default value in the flag definition.

pkg/compiler/compiler.go
```diff
@@ -46,15 +58,16 @@ func (c *Compiler) CompileTests() error {
 		return microerror.Mask(err)
 	}
 
-	name := harness.GetProjectName()
+	e2eBinary := harness.GetProjectName() + "-e2e"
+	e2eDir := filepath.Join(dir, c.testDir)
```
xh3b4sd: Test dir can now be `integration`, `integration/upgrade`, `2e2/foo`, or whatever users like.

.circleci/config.yml
```diff
@@ -46,6 +46,28 @@ jobs:
...
+    - checkout
+
+    - attach_workspace:
+        at: .
+
+    - run: ./e2e-harness localkube
+
+    - run: ./e2e-harness setup --remote=false
+
+    - run: ./e2e-harness test --test-dir=integration/sub-dir
```
fgimenez: Nice! I see that you have already added the wiring to the jobs so all good :)

I think it would be better to put the new test in a different directory, subdirs of `integration` were already covered.

xh3b4sd: Ack.

xh3b4sd:
🚀

PR #28, opened 2018-03-09
make integration/project.yaml optional

fgimenez:


pkg/project/project.go
```diff
@@ -186,7 +186,8 @@ func (p *Project) readProjectFile() (*E2e, error) {
 	}
 	projectFile := filepath.Join(dir, p.cfg.Dir, "project.yaml")
 	if _, err := os.Stat(projectFile); os.IsNotExist(err) {
-		return nil, microerror.Mask(err)
+		p.logger.Log("function", "readProjectFile", "level", "info", "message", "optional integration/project.yaml not found, continue build")
+		return &E2e{}, nil
```
xh3b4sd: Zero value of a pointer is usually `nil`.

pkg/project/project.go
```diff
@@ -186,7 +186,8 @@ func (p *Project) readProjectFile() (*E2e, error) {
 	}
 	projectFile := filepath.Join(dir, p.cfg.Dir, "project.yaml")
 	if _, err := os.Stat(projectFile); os.IsNotExist(err) {
-		return nil, microerror.Mask(err)
+		p.logger.Log("function", "readProjectFile", "level", "info", "message", "optional integration/project.yaml not found, continue build")
```
xh3b4sd: The path in the message is actually dynamically configured now. So it might be wrong in some cases.

pkg/project/project.go
fgimenez: In this case the value can be used down the road, no zero value here.

pkg/project/project.go
fgimenez: Good point, fixed.

PR #32, opened 2018-03-13
aligns code initialization with recent micrologger advances

xh3b4sd:


Gopkg.toml
```diff
@@ -52,6 +52,10 @@
   branch = "v2"
   name = "gopkg.in/yaml.v2"
 
+[[constraint]]
+  name = "k8s.io/api"
+  version = "kubernetes-1.8.5"
+
```
xh3b4sd: I had to pin this because of version conflicts. So far master was used which apparently became incompatible with the other pins we have.

cmd/init.go
```diff
@@ -23,9 +24,9 @@ func init() {
 }
 
 func runInit(cmd *cobra.Command, args []string) error {
-	logger, err := micrologger.New(micrologger.DefaultConfig())
+	logger, err := micrologger.New(micrologger.Config{})
 	if err != nil {
-		return err
+		return microerror.Mask(err)
```
xh3b4sd: Lets please always mask errors.

cmd/init.go
```diff
@@ -23,9 +24,9 @@ func init() {
 }
 
 func runInit(cmd *cobra.Command, args []string) error {
-	logger, err := micrologger.New(micrologger.DefaultConfig())
+	logger, err := micrologger.New(micrologger.Config{})
```
xh3b4sd: So this should actually be good enough.

xh3b4sd:
This is basically a vendor update. Rolling. Commented and pinged for awareness.

xh3b4sd:
Note that I had to remove the `Include administrators` in the branch protection settings here on Github in order to go ahead for vendor updates.

PR #37, opened 2018-03-16
increases wait time to 30 minutes again to prevent timeout errors

xh3b4sd:
This is to increase probability of e2e tests to succeed in combination with cloud providers plus being able to investigate on the machines in case things take way too long. I hope to find some things in the long run we can maybe improve or fix in our setup to make it more stable and reliable.

### DIFF
--- a/pkg/framework/wait.go
+++ b/pkg/framework/wait.go
@@ -8,7 +8,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


xh3b4sd:
I will probably do it spontaneously when I see something failing. I will give you a shout.

Issue #20, opened 2018-02-22
Error running tests because quay timeout getting a image 


pipo02mix:
Running e2e tests, I see from time to time this error. It comes from quay, we could add a retry to make it more robust

```
2018/02/22 09:50:04 Running command kubectl get certconfig
No resources found.
2018/02/22 09:50:04 Running command helm registry install quay.io/giantswarm/cert-resource-lab-chart:stable -- -n cert-resource-lab --set commonDomain=${COMMON_DOMAIN_GUEST} --set clusterName=${CLUSTER_NAME}
usage: appr [-h]
            {pull,run-server,show,inspect,list,delete-package,helm,version,logout,deploy,plugins,push,login,config,channel,jsonnet}
            ...
appr: error: 
message: 'NAME:   cert-resource-lab

  Error: context deadline exceeded
```

pipo02mix:
The problem was with the naming in AWS route table for one of our installations. It is solved

pipo02mix:
Closed the wrong one

fgimenez:
Failed again here with the same message https://circleci.com/gh/giantswarm/cert-operator/716

fgimenez:
Done in https://github.com/giantswarm/e2e-harness/pull/38

PR #38, opened 2018-03-20
Add retry to helm operations

fgimenez:
Towards https://github.com/giantswarm/e2e-harness/issues/20

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
 	giantclientset "github.com/giantswarm/apiextensions/pkg/clientset/versioned"
 	"github.com/giantswarm/microerror"
@@ -145,7 +146,13 @@ func (h *Host) InstallAwsOperator(values string) error {
 	if _, err := tmpfile.Write([]byte(awsOperatorChartValuesEnv)); err != nil {
 		return microerror.Mask(err)
 	}
-	if err := runCmd("helm registry install quay.io/giantswarm/aws-operator-chart@1.0.0-${CIRCLE_SHA1} -- -n aws-operator --values " + tmpfile.Name()); err != nil {
+
+	operation := func() error {
+		return HelmCmd("registry install quay.io/giantswarm/aws-operator-chart@1.0.0-${CIRCLE_SHA1} -- -n aws-operator --values " + tmpfile.Name())
+	}
+	notify := NewNotify("aws-operator-chart install")
+	err = backoff.RetryNotify(operation, NewCustomExponentialBackoff(), notify)
+	if err != nil {
 		return microerror.Mask(err)
 	}
 
@@ -157,7 +164,12 @@ func (h *Host) InstallCertOperator() error {
 	if err := ioutil.WriteFile(certOperatorValuesFile, []byte(certOperatorChartValuesEnv), os.ModePerm); err != nil {
 		return microerror.Mask(err)
 	}
-	if err := runCmd("helm registry install quay.io/giantswarm/cert-operator-chart:stable -- -n cert-operator --values " + certOperatorValuesFile); err != nil {
+	operation := func() error {
+		return HelmCmd("registry install quay.io/giantswarm/cert-operator-chart:stable -- -n cert-operator --values " + certOperatorValuesFile)
+	}
+	notify := NewNotify("cert-operator-chart install")
+	err := backoff.RetryNotify(operation, NewCustomExponentialBackoff(), notify)
+	if err != nil {
 		return microerror.Mask(err)
 	}
 
@@ -165,7 +177,11 @@ func (h *Host) InstallCertOperator() error {
 }
 
 func (h *Host) InstallCertResource() error {
-	err := runCmd("helm registry install quay.io/giantswarm/cert-resource-lab-chart:stable -- -n cert-resource-lab --set commonDomain=${COMMON_DOMAIN_GUEST} --set clusterName=${CLUSTER_NAME}")
+	operation := func() error {
+		return HelmCmd("registry install quay.io/giantswarm/cert-resource-lab-chart:stable -- -n cert-resource-lab --set commonDomain=${COMMON_DOMAIN_GUEST} --set clusterName=${CLUSTER_NAME}")
+	}
+	notify := NewNotify("cert-resource-lab-chart install")
+	err := backoff.RetryNotify(operation, NewCustomExponentialBackoff(), notify)
 	if err != nil {
 		return microerror.Mask(err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/project/project.go
```diff
@@ -110,9 +110,6 @@ func (p *Project) CommonSetupSteps() error {
 func (p *Project) CommonTearDownSteps() error {
 	p.logger.Log("info", "starting common teardown steps")
 	steps := []Step{
-		Step{
-			Run: "helm delete --purge sonobuoy-chart || true",
-		},
```
rossf7: For my interest why don't we need this anymore?

pkg/framework/host.go
```diff
@@ -145,7 +146,13 @@ func (h *Host) InstallAwsOperator(values string) error {
 	if _, err := tmpfile.Write([]byte(awsOperatorChartValuesEnv)); err != nil {
 		return microerror.Mask(err)
 	}
-	if err := runCmd("helm registry install quay.io/giantswarm/aws-operator-chart@1.0.0-${CIRCLE_SHA1} -- -n aws-operator --values " + tmpfile.Name()); err != nil {
+
+	operation := func() error {
+		return HelmCmd("registry install quay.io/giantswarm/aws-operator-chart@1.0.0-${CIRCLE_SHA1} -- -n aws-operator --values " + tmpfile.Name())
+	}
+	notify := NewNotify("aws-operator-chart install")
+	err = backoff.RetryNotify(operation, NewCustomExponentialBackoff(), notify)
```
xh3b4sd: We work towards abstraction and injection. I would like to configure the backoffs. Later we might want to write tests and then we can configure stop backoffs or whatever we need without refactoring it. Lets please keep things configurable.

pkg/project/project.go
fgimenez: Sorry, this change is not related to this PR at all, we removed sonobuoy support a long time ago from e2e-harness, we are getting `Error: release: "sonobuoy-chart" not found` on cleanup, like here https://circleci.com/gh/giantswarm/aws-operator/4498

pkg/framework/host.go
fgimenez: Sorry but i don't understand, could you please elaborate what do you mean by that? You can always pass an ad-hoc `&backoff.ExponentialBackOff` to `backoff.RetryNotify` if you want to customize further, `NewCustomExponentialBackoff` and `NewNotify` just give defaults.

pkg/project/project.go
rossf7: Cool thanks for the explanation. Good to remove this if it's not used.

pkg/framework/host.go
xh3b4sd: We try to improve the framework here. I already did a lot of restructuring and refactoring. The more we abstract and make configurable the more testable becomes the code. I want to get into a situation in which we might be able to write tests for our code. At the moment this is not really possible because we assume a lot based on the global scope and process environment and so on. When something like the backoff and helm client is abstracted and configurable we can easily inject mocks and write unit tests or something. All this is just towards a better design and adding code which assumes more based on the global scope or process environment and hard code stuff is not making it better here.

fgimenez: Aha, you meant injecting the backoff, ok done PTAL.

xh3b4sd: Yeah yeah, the usual dependency injection thing we do.

pkg/framework/host.go
```diff
@@ -222,20 +242,20 @@ func (h *Host) PodName(namespace, labelSelector string) (string, error) {
 	return pod.Name, nil
 }
 
-func (h *Host) Setup() error {
+func (h *Host) Setup(b *backoff.ExponentialBackOff) error {
```
xh3b4sd: I would just have made the backoff configurable per `Host` type initialization with the config structure. Then the interfaces do not change. I think in almost all cases we do not need to configure a different backoff per framework per action.

fgimenez: Sure, done PTAL.

pkg/framework/host.go
```diff
@@ -63,9 +64,10 @@ Installation:
 type Host struct {
 	g8sClient *giantclientset.Clientset
 	k8sClient kubernetes.Interface
+	backoff   *backoff.ExponentialBackOff
 }
 
-func NewHost() (*Host, error) {
+func NewHost(b *backoff.ExponentialBackOff) (*Host, error) {
```
xh3b4sd: We have a specific pattern in most places for these kind of things. Is there a reason why we should not apply it here? Like having a config structure. I do not see why this here should be different. The next guy has to add something and then the interface changes again and somebody has to do a lot of refactoring again just to get some other dependency in here only because we always only take the quick and dirty approach. The next thing we want to configure here is the helm/appr clients. So this is foreseeable already.

pkg/framework/host.go
```diff
@@ -83,6 +85,7 @@ func NewHost() (*Host, error) {
 	h := &Host{
 		g8sClient: g8sClient,
 		k8sClient: k8sClient,
+		backoff:   b,
```
xh3b4sd: It would be cool to consider ordering as we do in other places.

pkg/framework/host.go
fgimenez: Done PTAL.

xh3b4sd (approved):
Thank you for your energy here.

pkg/framework/retry.go
```diff
@@ -0,0 +1,35 @@
...
+		Multiplier:          backoff.DefaultMultiplier,
+		MaxInterval:         backoff.DefaultMaxInterval,
+		MaxElapsedTime:      maxElapsedTime,
+		Clock:               backoff.SystemClock,
+	}
+
+	b.Reset()
+
+	return b
+}
```
kopiczko: Is it used anywhere?

pkg/framework/retry.go
```diff
@@ -0,0 +1,35 @@
...
+
+func NewNotify(operationName string) func(error, time.Duration) {
+	return func(err error, delay time.Duration) {
+		log.Printf(fmt.Sprintf("%s failed, retrying with delay %.0fm%.0fs: '%#v'", operationName, delay.Minutes(), delay.Seconds(), err))
+	}
+}
```
kopiczko: It doesn't need to be exported. I'm big on making everything private by default :)

pkg/framework/host.go
```diff
@@ -61,11 +62,20 @@ Installation:
...
 
-func NewHost() (*Host, error) {
+type Config struct {
+	Backoff *backoff.ExponentialBackOff
+}
+
+func NewHost(c *Config) (*Host, error) {
+	if c.Backoff == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Backoff must not be empty", c)
+	}
```
kopiczko: You could use a default one here instead of failing.

pkg/framework/retry.go
fgimenez: I was planning to use it from the consumers, using it to set a default value when missing on initialization is much nicer :) Also moved to private.

pkg/framework/retry.go
fgimenez: :100:

pkg/framework/host.go
fgimenez: Nice, done.

kopiczko (approved):
<3

PR #40, opened 2018-03-23
aligns framework config with usual approach

xh3b4sd:


pkg/framework/host.go
```diff
@@ -61,17 +61,17 @@ Installation:
 `
 )
 
+type HostConfig struct {
```
xh3b4sd: There is `Guest` and `Host` in this package. This is also not about the framework config but about the host config.

kopiczko (approved):

pkg/framework/host.go
```diff
@@ -61,17 +61,17 @@ Installation:
...
 	g8sClient *giantclientset.Clientset
 	k8sClient kubernetes.Interface
 }
 
-type Config struct {
-	Backoff *backoff.ExponentialBackOff
-}
-
-func NewHost(c *Config) (*Host, error) {
+func NewHost(c HostConfig) (*Host, error) {
```
xh3b4sd: I do not know why this should be a pointer. We are not doing it like this. Nowhere.

kopiczko: AFAIK changing it to a value can't brake anything in this case.

PR #42, opened 2018-04-03
Retry setup operations

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2571

This can prevent transient errors during setup by retrying the failing tasks.

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -2,10 +2,12 @@ package project
 
 import (
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	"github.com/spf13/afero"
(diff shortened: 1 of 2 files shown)


pkg/project/project.go
```diff
@@ -98,8 +100,18 @@ func (p *Project) CommonSetupSteps() error {
 			},
 		}}
 
+	newNotify := func(operationName string) func(error, time.Duration) {
+		return func(err error, delay time.Duration) {
+			log.Printf(fmt.Sprintf("%s failed, retrying with delay %.0fm%.0fs: '%#v'", operationName, delay.Minutes(), delay.Seconds(), err))
```
xh3b4sd: The project here has a logger. Lets please use it. The duration we get implements stringer. Lets use this as well.

pkg/project/project.go
```diff
@@ -98,8 +100,18 @@ func (p *Project) CommonSetupSteps() error {
...
+		}
+	}
 	for _, s := range steps {
-		if err := p.RunStep(s); err != nil {
+		operation := func() error {
+			return p.RunStep(s)
```
xh3b4sd: Lets please use explicit error masking.

fgimenez: Sure, done

pkg/project/project.go
fgimenez: Good point, using the logger member. About the duration, you mean using the `%s` verb with it as a whole, without treating minutes and seconds separately?

xh3b4sd: Yeah, we can also just call `String()` so the formatter does not have to do it, but this is minor. My main point is that the formatting done by the stringer is simpler to use and its output easier to read because it omits minutes and hours if the actual duration is just about seconds.

fgimenez: Makes sense, done.

xh3b4sd (approved):

indyfree (approved):
LGTM

PR #43, opened 2018-04-04
Add verbose flag to test binary execution

fgimenez:
Closes https://github.com/giantswarm/e2e-harness/issues/30

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -150,9 +150,9 @@ func (p *Project) CommonTearDownSteps() error {
 func (p *Project) Test() error {
 	p.logger.Log("info", "started tests")
 
-	// ./e2e is mounted in /e2e in the test container, and the binary with the e2e
-	// tests is named <project_name>-e2e so the final location in the test container
-	// is /e2e/<project_name>-e2e
+	// --test-dir is mounted in /e2e in the test container, and the binary with
+	// the e2e tests is named <project_name>-e2e so the final location in the
+	// test container is /e2e/<project_name>-e2e
 	name := harness.GetProjectName()
 	binaryPath := fmt.Sprintf("/e2e/%s-e2e", name)
 
@@ -161,7 +161,7 @@ func (p *Project) Test() error {
 		return microerror.Mask(err)
 	}
 
-	if err := p.runner.RunPortForward(os.Stdout, binaryPath, e2e.Test.Env...); err != nil {
+	if err := p.runner.RunPortForward(os.Stdout, binaryPath+" -test.v", e2e.Test.Env...); err != nil {
 		return microerror.Mask(err)
 	}
 


pkg/project/project.go
```diff
@@ -161,7 +161,7 @@ func (p *Project) Test() error {
 		return microerror.Mask(err)
 	}
 
-	if err := p.runner.RunPortForward(os.Stdout, binaryPath, e2e.Test.Env...); err != nil {
+	if err := p.runner.RunPortForward(os.Stdout, binaryPath+" -test.v", e2e.Test.Env...); err != nil {
```
xh3b4sd: I don't follow here. What is `-test.v`?  This looks like the binary path is now broken.

fgimenez: It is the flag to show verbose output when running the test binary. Instead of just the binary path `/path/to/test-binary`, we now pass the binary path plus the verbose flag `/path/to/binary -test.v` to the runner method. See how the test results look like with these changes https://circleci.com/gh/giantswarm/e2e-harness/619

xh3b4sd: I see that flag the very first time. :D Ok cool. Looks like it does what it should.

xh3b4sd (approved):

pkg/project/project.go
fgimenez: yep :) flags are a little bit different for compiled test binaries:
```
$ ./integration/e2e-harness-e2e --help
Usage of integration/e2e-harness-e2e:
  -alsologtostderr
    	log to standard error as well as files
  -log_backtrace_at value
    	when logging hits line file:N, emit a stack trace
  -log_dir string
    	If non-empty, write log files in this directory
  -logtostderr
    	log to standard error instead of files
  -stderrthreshold value
    	logs at or above this threshold go to stderr
  -test.bench regexp
    	run only benchmarks matching regexp
  -test.benchmem
    	print memory allocations for benchmarks
  -test.benchtime d
    	run each benchmark for duration d (default 1s)
  -test.blockprofile file
    	write a goroutine blocking profile to file
  -test.blockprofilerate rate
    	set blocking profile rate (see runtime.SetBlockProfileRate) (default 1)
  -test.count n
    	run tests and benchmarks n times (default 1)
  -test.coverprofile file
    	write a coverage profile to file
  -test.cpu list
    	comma-separated list of cpu counts to run each test with
  -test.cpuprofile file
    	write a cpu profile to file
  -test.failfast
    	do not start new tests after the first test failure
  -test.list regexp
    	list tests, examples, and benchmarks matching regexp then exit
  -test.memprofile file
    	write a memory profile to file
  -test.memprofilerate rate
    	set memory profiling rate (see runtime.MemProfileRate)
  -test.mutexprofile string
    	write a mutex contention profile to the named file after execution
  -test.mutexprofilefraction int
    	if >= 0, calls runtime.SetMutexProfileFraction() (default 1)
  -test.outputdir dir
    	write profiles to dir
  -test.parallel n
    	run at most n tests in parallel (default 8)
  -test.run regexp
    	run only tests and examples matching regexp
  -test.short
    	run smaller test suite to save time
  -test.testlogfile file
    	write test action log to file (for use only by cmd/go)
  -test.timeout d
    	panic test binary after duration d (default 0, timeout disabled)
  -test.trace file
    	write an execution trace to file
  -test.v
    	verbose: print additional output
  -v value
    	log level for V logs
  -vmodule value
    	comma-separated list of pattern=N settings for file-filtered logging

```

Issue #25, opened 2018-03-09
Refactor AWS specific methods in the framework package


fgimenez:

kopiczko:
Outdated.

PR #45, opened 2018-04-04
Refactor install operator methods

fgimenez:
Towards https://github.com/giantswarm/e2e-harness/issues/25

Added generic methods for installing operators, both stable and from branch versions.

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -144,46 +144,51 @@ func (h *Host) DeleteGuestCluster() error {
 	return h.WaitForPodLog("giantswarm", logEntry, operatorPodName)
 }
 
-func (h *Host) InstallAwsOperator(values string) error {
-	awsOperatorChartValuesEnv := os.ExpandEnv(values)
+func (h *Host) InstallStableOperator(name, cr, values string) error {
+	err := h.installOperator(name, cr, values, ":stable")
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	return nil
+}
 
-	tmpfile, err := ioutil.TempFile("", "aws-operator-values")
+func (h *Host) InstallBranchOperator(name, cr, values string) error {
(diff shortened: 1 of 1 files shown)


pipo02mix (approved):
lgmt

pkg/framework/host.go
```diff
@@ -144,46 +144,51 @@ func (h *Host) DeleteGuestCluster() error {
...
 		return microerror.Mask(err)
 	}
 	defer os.Remove(tmpfile.Name())
 
-	if _, err := tmpfile.Write([]byte(awsOperatorChartValuesEnv)); err != nil {
+	if _, err := tmpfile.Write([]byte(chartValuesEnv)); err != nil {
 		return microerror.Mask(err)
 	}
 
+	cmd := fmt.Sprintf("registry install quay.io/giantswarm/%[1]s-chart%[2]s -- -n %[1]s --values %[3]s", name, version, tmpfile.Name())
```
xh3b4sd: How does the formatting with the square brackets work? Never saw this before.

kopiczko: I also had to educate myself: https://golang.org/pkg/fmt/ -> search for `Explicit argument indexes`.

indyfree (approved):
Nice

pkg/framework/host.go
fgimenez: Exactly, here, given that the name is used twice in the format string, the goal is not having to repeat values, see https://play.golang.org/p/IfqFTpiV5eh for example in contrast to https://play.golang.org/p/gnzXPMw2J5F

fgimenez:
I removed the now unused cert-operator constants too. Going to merge now, there is still some aws specific code related to the CR chart and patching mainly, will work on that next as part of https://github.com/giantswarm/e2e-harness/issues/25. We can iterate on the operator code changes in this PR if needed.

PR #47, opened 2018-04-05
Expose InstallOperator

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1902

chart-operator doesn't follow the version pattern `1.0.0-${CIRCLE_SHA1}` we have in other operators, because it's going to be self-installed it needs to follow the new pattern with actual semantic versions (like `3.1.7`) and stability-related channels. Because of that the currently exposed methods for installing operators (from the `stable` channel or from `1.0.0-${CIRCLE_SHA1}` version) are not enough, we need to specify exactly the channel to pull the chart from (in PRs this will be `CIRCLE_SHA1`). This can be done exposing the base method `installOperator`.

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -115,7 +115,7 @@ func (h *Host) DeleteGuestCluster() error {
 }
 
 func (h *Host) InstallStableOperator(name, cr, values string) error {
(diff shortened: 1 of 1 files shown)


rossf7 (approved):
LGTM. Nice enhancement for chart-operator support.

PR #48, opened 2018-04-06
refactor DeleteGuestCluster

TheoBrigitte:
Basically passing all needed values as arguments.

Towards not being aws specific.

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -99,17 +99,15 @@ func (h *Host) AWSCluster(name string) (*v1alpha1.AWSConfig, error) {
 	return cluster, nil
 }
 
-func (h *Host) DeleteGuestCluster() error {
-	if err := runCmd("kubectl delete awsconfig ${CLUSTER_NAME}"); err != nil {
+func (h *Host) DeleteGuestCluster(cr, name, logEntry string) error {
+	if err := runCmd(fmt.Sprintf("kubectl delete %s ${CLUSTER_NAME}", cr)); err != nil {
 		return microerror.Mask(err)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/framework/host.go
```diff
@@ -99,18 +99,16 @@ func (h *Host) AWSCluster(name string) (*v1alpha1.AWSConfig, error) {
 	return cluster, nil
 }
 
-func (h *Host) DeleteGuestCluster() error {
-	if err := runCmd("kubectl delete awsconfig ${CLUSTER_NAME}"); err != nil {
+func (h *Host) DeleteGuestCluster(cr, name, logEntry string) error {
```
fgimenez: It would be nice to align with the current Install methods and have the name parameter before the cr, see https://github.com/giantswarm/e2e-harness/blob/master/pkg/framework/host.go#L117

fgimenez (approved):
LGTM, nit about param order.

kopiczko (approved):
LGTM when @fgimenez is happy

xh3b4sd:
FYI when you do things like this which are not backwards compatible it would be very nice to update the projects which depend on this. I need to figure this out now on my own and do yak shaving.

TheoBrigitte:
Sorry, I thought about this for a sec and forgot.
For aws you would actually need to replace `DeleteGuestCluster` with `DeleteGuestCluster("awsconfig", "aws-operator", "deleted the guest cluster main stack")`

xh3b4sd:
Almost right. See https://github.com/giantswarm/e2e-harness/blob/00dce95dd9b359cf782f79128961d45239c915e8/pkg/framework/host.go#L105. The name is the first parameter.

xh3b4sd:
Did not check who changed this after you.

PR #49, opened 2018-04-10
provide generic method to install resource

TheoBrigitte:
As mentioned in giantswarm/azure-operator#162 (comment)
I implemented this method toward having a generic way to install resource.

But I am not sure how to handle the different wait cases here.
One solution would be to pass the wait function as an argument and if present call `waitFor(argFunction())`
It would then also be nice to export `h.secret` and `h.runningPod`.

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -162,7 +162,7 @@ func (h *Host) InstallOperator(name, cr, values, version string) error {
 	return nil
 }
 
-func (h *Host) InstallResource(name, values string) error {
+func (h *Host) InstallResource(name, values string, conditions ...func() error) error {
 	chartValuesEnv := os.ExpandEnv(values)
 
 	tmpfile, err := ioutil.TempFile("", name+"-values")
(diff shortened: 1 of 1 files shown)


TheoBrigitte:
As mentioned in https://github.com/giantswarm/azure-operator/pull/162#discussion_r180112389
I implemented this method toward having a generic way to install resource.

But I am not sure how to handle the different wait cases here.
One solution would be to pass the wait function as an argument and if present call `waitFor(argFunction())`
It would then also be nice to export `h.secret` and `h.runningPod`.

TheoBrigitte:
This is even better, so we can wait for multiple resources or just omit this parameter. I would like to have this implemented in this PR, so we have a strong `InstallResource` to rely on.

I realised that `InstallOperator` is now a super set of `InstallResource`. But let's have that in another PR along with exporting the different function to pass to `waitFor`.

fgimenez:
Nice! agree, better refactor `InstallOperator` on a different PR.

PR #50, opened 2018-04-10
exposes initialized k8s clients for guest and host frameworks

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2737. The idea is here to let us use the k8s clients freely during development. Once we notice functionality becomes common ground we can move it to the framework but for prototyping and playing around it is super useful to have the clients setup and available in integration tests.

### DIFF
--- a/pkg/framework/guest.go
+++ b/pkg/framework/guest.go
@@ -22,12 +22,14 @@ const (
 )
 
 type Guest struct {
-	k8sClient kubernetes.Interface
+	k8sClient  kubernetes.Interface
+	restConfig *rest.Config
 }
 
 func NewGuest() (*Guest, error) {
 	g := &Guest{
-		k8sClient: nil,
+		k8sClient:  nil,
+		restConfig: nil,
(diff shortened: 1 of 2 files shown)


fgimenez (approved):
LGTM

xh3b4sd:
Note that I also exposed the rest config being created by the frameworks. We can probably drop this at some point in the future but for now I want to be free to experiment a little bit in integration tests to see how it can work when dealing with host and gest clusters within the same integration test.

kopiczko (approved):

fgimenez:
:+1: with the config exposed this can be simplified https://github.com/giantswarm/helmclient/blob/master/integration/test/basic/basic_test.go#L23-L30

xh3b4sd:
Yeah totally. I am thinking to maybe even create the appr and helm clients with the frameworks setup internally. Then we can drop this at all but lets see where this goes. Lets get experiences first and iterate from here.

PR #52, opened 2018-04-12
Always rebuild test binaries 

MarcelMue:
This has been annoying me when testing locally, I always forget to manually remove the `x-e2e` binary and end up running my old stuff.

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -95,7 +95,8 @@ func runTest(cmd *cobra.Command, args []string) error {
 		c := compiler.Config{
 			Logger: logger,
 
-			TestDir: testCmdTestDir,
+			RemoteCluster: cfg.RemoteCluster,
+			TestDir:       testCmdTestDir,
 		}
 
 		comp = compiler.New(c)
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -95,8 +95,7 @@ func runTest(cmd *cobra.Command, args []string) error {
 		c := compiler.Config{
 			Logger: logger,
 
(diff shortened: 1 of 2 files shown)


cmd/test.go
```diff
@@ -95,7 +95,8 @@ func runTest(cmd *cobra.Command, args []string) error {
 		c := compiler.Config{
 			Logger: logger,
 
-			TestDir: testCmdTestDir,
+			RemoteCluster: cfg.RemoteCluster,
```
MarcelMue: Not sure if this is correct to use here. I thought that `RemoteCluster` is always false when building locally and true when doing it on Circle. If this is wrong, then I'll introduce a new flag.

MarcelMue: No this is wrong, we also use it elsewhere.

fgimenez: Yes, `RemoteCluster` refers to where the test cluster lives; you can run your test locally (building binaries from your disk) and use a remote cluster for running the tests.

fgimenez (commented):
Thanks a lot for this! Really annoying, yes :) I would go for always building the test binary. The only reason for not building the main binary is that, in CI, it has been built before in a previous job by architect, and we can save some time if we don't build it if present. But the e2e binary should always be rebuilt IMO, WDYT?

MarcelMue:
Yes, that makes way more sense. I just removed the check for the test binary now, this was meant to speed up circle it's never built twice in circle. So this should be fine.

fgimenez (approved):
<3

PR #56, opened 2018-04-16
uses backoffs for retry logic in guest framework

xh3b4sd:
This PR fixes an issue with the retry logic. Note there are still other issues but for now we only focus on a subset within this PR. See also https://gigantic.slack.com/archives/C3C7ZQXC1/p1523897705000242.

### DIFF
--- a/pkg/framework/backoff.go
+++ b/pkg/framework/backoff.go
@@ -1,0 +1,32 @@
+package framework
+
+import (
+	"time"
+
+	"github.com/cenkalti/backoff"
+)
+
+const (
+	LongMaxWait  = 30 * time.Minute
+	ShortMaxWait = 2 * time.Minute
+)
+
+const (
+	LongMaxInterval  = 60 * time.Second
+	ShortMaxInterval = 5 * time.Second
+)
+
+func newExponentialBackoff(maxWait, maxInterval time.Duration) *backoff.ExponentialBackOff {
+	b := &backoff.ExponentialBackOff{
+		InitialInterval:     backoff.DefaultInitialInterval,
+		RandomizationFactor: backoff.DefaultRandomizationFactor,
+		Multiplier:          backoff.DefaultMultiplier,
+		MaxInterval:         maxInterval,
+		MaxElapsedTime:      maxWait,
+		Clock:               backoff.SystemClock,
+	}
+
+	b.Reset()
+
+	return b
+}
--- a/pkg/framework/error.go
+++ b/pkg/framework/error.go
@@ -30,9 +30,16 @@ func IsUnexpectedStatusPhase(err error) bool {
 	return microerror.Cause(err) == unexpectedStatusPhaseError
 }
 
-var waitTimeoutError = microerror.New("waitTimeout")
+var waitError = microerror.New("wait")
 
-// IsWaitTimeout asserts invalidConfigError.
+// IsWait asserts waitError.
+func IsWait(err error) bool {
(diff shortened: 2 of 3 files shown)


pkg/framework/backoff.go
```diff
@@ -0,0 +1,32 @@
...
+	LongMaxWait  = 30 * time.Minute
+	ShortMaxWait = 2 * time.Minute
+)
+
+const (
+	LongMaxInterval  = 60 * time.Second
+	ShortMaxInterval = 5 * time.Second
+)
+
+func newExponentialBackoff(maxWait, maxInterval time.Duration) *backoff.ExponentialBackOff {
```
xh3b4sd: We start using a backoff here for the retry logic. Right now retrying and timing out is kind of home grown and a bit mixed up with the usage of backoffs. Here we try to get this a bit more straight to make the code easier and prevent nasty side effects.

pkg/framework/error.go
```diff
@@ -30,9 +30,16 @@ func IsUnexpectedStatusPhase(err error) bool {
 	return microerror.Cause(err) == unexpectedStatusPhaseError
 }
 
-var waitTimeoutError = microerror.New("waitTimeout")
+var waitError = microerror.New("wait")
```
xh3b4sd: Just adding a common wait error which we annotate for more information.

pkg/framework/guest.go
```diff
@@ -105,32 +106,63 @@ func (g *Guest) Setup() error {
...
 func (g *Guest) WaitForAPIDown() error {
-	apiDown := func() error {
-		_, err := g.k8sClient.
-			CoreV1().
-			Services("default").
-			Get("kubernetes", metav1.GetOptions{})
-
-		if err == nil {
-			return microerror.Mask(fmt.Errorf("API up"))
+	time.Sleep(1 * time.Second)
```
xh3b4sd: I just added this arbitrary wait second to prevent cases in which the API is still available and this the check below succeeds unexpectedly.

pkg/framework/guest.go
```diff
@@ -105,32 +106,63 @@ func (g *Guest) Setup() error {
...
+
+	log.Println("level", "debug", "message", "waiting for k8s API to be down")
```
xh3b4sd: Note that I add logging here via a native lib. The output is n structured JSON but I want to introduce an injected micrologger in the next PRs. These should then be quite nice and easy.

pkg/framework/guest.go
```diff
@@ -105,32 +106,63 @@ func (g *Guest) Setup() error {
...
+
+	o := func() error {
+		_, err := g.k8sClient.CoreV1().Services("default").Get("kubernetes", metav1.GetOptions{})
+		if err != nil {
```
MarcelMue: You could check for the actual error you expect here, that way the 1 second wait isn't needed. This would also prevent this from passing because something is actually wrong with the API, or is that intended?

MarcelMue (commented):
Small comments, looks good otherwise.

pkg/framework/guest.go
xh3b4sd: This is actually how it was before just the other way around. When you tell me which error to match against then we can improve this. The thing is that there is nothing like a "this thing is not up" error. It might be an url error, or net error or timeout error or whatever.

MarcelMue: I see the issue, still not very satisfying I guess.

MarcelMue (approved):

PR #57, opened 2018-04-17
makes common backoff creation publicly available

xh3b4sd:
This is to be able to reuse the common backoff functionality in our operators where necessary.

### DIFF
--- a/pkg/framework/backoff.go
+++ b/pkg/framework/backoff.go
@@ -16,7 +16,7 @@ const (
 	ShortMaxInterval = 5 * time.Second
 )
 
-func newExponentialBackoff(maxWait, maxInterval time.Duration) *backoff.ExponentialBackOff {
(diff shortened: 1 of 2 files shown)


indyfree (approved):
LGTM

pkg/framework/backoff.go
```diff
@@ -16,7 +16,7 @@ const (
 	ShortMaxInterval = 5 * time.Second
 )
 
-func newExponentialBackoff(maxWait, maxInterval time.Duration) *backoff.ExponentialBackOff {
+func NewExponentialBackoff(maxWait, maxInterval time.Duration) *backoff.ExponentialBackOff {
```
kopiczko: We can decide it will be other type of backoff so maybe it makes sense to rename it to more generic form like `NewBackOff`.

Also nitpick, capital `O` in BackOff.

PR #58, opened 2018-04-17
injects micrologger

xh3b4sd:
Following up on https://github.com/giantswarm/e2e-harness/pull/56 to inject the logger in the guest framework now.

### DIFF
--- a/pkg/framework/guest.go
+++ b/pkg/framework/guest.go
@@ -1,12 +1,12 @@
 package framework
 
 import (
-	"log"
 	"os"
 	"time"
 
 	"github.com/cenkalti/backoff"
 	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


pkg/framework/guest.go
```diff
@@ -108,7 +120,7 @@ func (g *Guest) Setup() error {
 func (g *Guest) WaitForAPIDown() error {
 	time.Sleep(1 * time.Second)
 
-	log.Println("level", "debug", "message", "waiting for k8s API to be down")
+	g.logger.Log("level", "debug", "message", "waiting for k8s API to be down")
```
kopiczko: Does that mean we'll have JSON in circle output?

kopiczko (approved):
It is how it is. Let's roll.

pkg/framework/guest.go
xh3b4sd: I do not understand the problem. It is configurable. You decide what you do in your code. When you want to do something else you change your code.

kopiczko: To change to output, you have to manipulate the code, instead out having a flag like `--micrologger-output=json`.

Second thing is that when you want to change something you need make sure every logger instance in the code is changed, because we always want all the loggers to be aligned. That is a huge chore, and I don't see benefits.

xh3b4sd: When you want to change it everywhere you change the default in micrologger. Come on Pawel. What is the problem?

kopiczko: The problem is changing it per project basis. I don't want to change the default as this is good default most cases.

xh3b4sd: This is not a problem. We can just do it if we want.

kopiczko: We can. It's unnecessarily difficult.

xh3b4sd: The downsides and side effects we would have with anything globally configured would just be worse. That does not make any sense to me.

PR #59, opened 2018-04-17
purges releases before installing them

xh3b4sd:
We saw lots of issues recently where retries caused helm commands to fail. This PR tries to help with these problems.

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -197,7 +197,17 @@ func (h *Host) InstallResource(name, values string, conditions ...func() error) 
 
 func (h *Host) InstallCertResource() error {
 	operation := func() error {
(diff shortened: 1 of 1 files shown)


fgimenez (approved):
Nice thx, there's an error in CI on image build, probably ignorable (it just failed in one of the jobs, the image was properly built in the rest of them).

xh3b4sd:
Why does it fail though? Can we improve on this? I do not see what went wrong. Can we improve logging and retry or something?

indyfree (approved):
LGTM if CI is not a problem.

xh3b4sd:
Rerun, green now.

fgimenez:
> Why does it fail though? Can we improve on this? I do not see what went wrong. Can we improve logging and retry or something?

Sure, I've proposed #60 to show the eventual build errors.

PR #61, opened 2018-04-18
refactor InstallOperator to make use of InstallResource

TheoBrigitte:
Follow-up from https://github.com/giantswarm/e2e-harness/pull/49

After providing a more generic method to install chart resources, I realised that `InstallOperator` was nearly the same as the new `InstallResource` method.

Here is a refactoring of `InstallOperator` which is now just using `InstallResource`.

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -132,34 +132,8 @@ func (h *Host) InstallBranchOperator(name, cr, values string) error {
 }
 
 func (h *Host) InstallOperator(name, cr, values, version string) error {
-	chartValuesEnv := os.ExpandEnv(values)
-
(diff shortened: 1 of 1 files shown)


MarcelMue:
Code checkout failed on Circle? Looks like outage or something 🤷‍♂️

xh3b4sd (commented):
When we refactor this here it would be the better way to use our appr and helm client libs and stop using the shell out magic.

TheoBrigitte:
CI is happy, are we happy?

fgimenez (approved):
LGTM, also agree with the refactor, but as part of other PR i guess.

indyfree (approved):
LGMT. Agree to refactor to use client libs in other PR

PR #63, opened 2018-04-18
improves waiting for secret by using backoff

xh3b4sd:


pkg/framework/host.go
```diff
@@ -207,28 +207,60 @@ func (h *Host) InstallResource(name, values string, conditions ...func() error)
 }
 
 func (h *Host) InstallCertResource() error {
-	operation := func() error {
-		// NOTE we ignore errors here because we cannot get really useful error
-		// handling done. This here should anyway only be a quick fix until we use
-		// the helm client lib. Then error handling will be better.
-		HelmCmd("delete --purge cert-resource-lab")
+	{
```
xh3b4sd: The diff is weird because I put this into a separate block.

pkg/framework/host.go
```diff
@@ -207,28 +207,60 @@ func (h *Host) InstallResource(name, values string, conditions ...func() error)
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		log.Println("level", "debug", "message", "k8s secret is there")
 	}
 
-	secretName := fmt.Sprintf("%s-api", os.Getenv("CLUSTER_NAME"))
-	log.Printf("waiting for secret %v\n", secretName)
-	return waitFor(h.secret("default", secretName))
```
xh3b4sd: `waitFor` got replaced by using the backoff like in the guest framework. I want to get this straight for the host framework as well. We already noticed the wait implementation is buggy in certain cases.

pkg/framework/host.go
```diff
@@ -207,28 +207,60 @@ func (h *Host) InstallResource(name, values string, conditions ...func() error)
...
+
+	{
+		log.Println("level", "debug", "message", "waiting for k8s secret to be there")
+
+		o := func() error {
+			n := fmt.Sprintf("%s-api", os.Getenv("CLUSTER_NAME"))
+			_, err := h.k8sClient.CoreV1().Secrets("default").Get(n, metav1.GetOptions{})
+			if err != nil {
+				// TODO remove this when not needed for debugging anymore
+				fmt.Printf("%#v\n", err)
```
xh3b4sd: I want to keep this for some time. There was a failing build because the secret was never there. I hope to catch this situation in the next days again and then get more information. Don't know how to reproduce this in Circle so we need to have luck.

pkg/framework/host.go
```diff
@@ -207,28 +207,60 @@ func (h *Host) InstallResource(name, values string, conditions ...func() error)
...
+		}
+		b := NewExponentialBackoff(ShortMaxWait, ShortMaxInterval)
+		n := newNotify("cert-resource-lab-chart install")
+		err := backoff.RetryNotify(o, b, n)
 		if err != nil {
 			return microerror.Mask(err)
 		}
 
-		return nil
+		log.Println("level", "debug", "message", "installed cert resource chart")
```
xh3b4sd: Doing the same logging dance like in the guest framework. Later we can just inject a micrologger.

pkg/framework/host.go
```diff
@@ -207,28 +207,60 @@ func (h *Host) InstallResource(name, values string, conditions ...func() error)
...
 	}
-	notify := newNotify("cert-resource-lab-chart install")
-	err := backoff.RetryNotify(operation, h.backoff, notify)
-	if err != nil {
-		return microerror.Mask(err)
+
+	{
+		log.Println("level", "debug", "message", "waiting for k8s secret to be there")
+
+		o := func() error {
```
TheoBrigitte: Why not using the h.secret() function here? It is doing the same thing.

```
n := fmt.Sprintf("%s-api", os.Getenv("CLUSTER_NAME"))
o := h.secret("default", n)
```

xh3b4sd: Because I want to refactor the code and get the mental model straight in my head instead of using the old stuff that is there. When it turns out we want to have the operation being separated and reusable I will be glad to provide a PR to do this. But I need to see this and get a feel for the framework first.

fgimenez (approved):
LGTM, i think we should change `waitFor` asap if it's not working well, other methods rely on it.

xh3b4sd:
My thinking was to replace it with the backoffs. I did this in the guest framework already. Started in the host framework now. Will continue with this. Later cleanup.

PR #66, opened 2018-04-24
enable network management from container

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1902

We want to be able to test chart-operator bootstrap in cluster-operator without having to spin up a guest cluster, and instead check the creation on the "host cluster" (circleci machine in e2e CI runs, minikube for local runs).

In order to do that we need:
* Redirect calls to domains like `api.${CLUSTER_NAME}.aws.gigantic.io` to `127.0.0.1`
* Forward requests on `127.0.0.1` port `80` (default for guest k8s API) to port `8443` (default for localkube/minikube).

These changes allow both requirements by passing `--add-host` parameters to the test container (first requirement) and setting up iptables in the test container for the unprivileged `e2e-harness` user (second requirement).

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -90,7 +90,9 @@ func (d *Docker) baseRun(out io.Writer, entrypoint string, args []string, env ..
 	// host1:ip1,host2:ip2...
 	addHostEntries := strings.Split(os.Getenv("ADD_HOSTS"), ",")
 	for _, entry := range addHostEntries {
-		baseArgs = append(baseArgs, "--add-host", entry)
+		if entry != "" {
+			baseArgs = append(baseArgs, "--add-host", entry)
+		}
 	}
 
 	// add environment variables
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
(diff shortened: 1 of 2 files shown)


pkg/docker/docker.go
```diff
@@ -82,6 +83,16 @@ func (d *Docker) baseRun(out io.Writer, entrypoint string, args []string, env ..
...
+	}
+
+	// add-host entries, the ADD_HOSTS env var will look like:
+	// host1:ip1,host2:ip2...
+	addHostEntries := strings.Split(os.Getenv("ADD_HOSTS"), ",")
+	for _, entry := range addHostEntries {
+		if entry != "" {
+			baseArgs = append(baseArgs, "--add-host", entry)
+		}
 	}
```
kopiczko: How iptables are set?

fgimenez: iptables are installed in the image and the `e2e-harness` user (the default) is allowed to call it as root. It's up to the consumer projects to use the command as needed.

xh3b4sd (commented):
IMO we should work towards using go libraries to have more control over the process and its environment in the long run. This here is all just a quick fix and at the end somebody has to do some intensive yak shaving to refactor all of this in future stories. Lets please not always take the shortcuts. When you think this is right then lets go, but I will keep pushing on this level.

kopiczko (approved):
LGTM. I'm not sure how we can do it with pure Go in general.

Do you suggest editing /etc/hosts? How to allow rewriting iptables rules?

fgimenez:
For the record, my initial approach for adding DNS entries was to change `/etc/hosts` but all the attempts failed (editing directly, hard linking and bind mounting). In this case I think we should take advantage of `--add-host` one way or another.

About iptables I have the same question as @kopiczko, don't know how we can write rules from code, other than calling the cli utility.

xh3b4sd:
See http://bfy.tw/Ho6v.

fgimenez:
Nice, thx! Not sure if that would work in e2e-harness container, we have a non-priviledged user, and probably this library requires root. Anyway, good to know, anyone is free to try it in consumer projects.

kopiczko:
This is from top result readme:

> go-iptables wraps invocation of iptables utility with functions to append and delete rules; create, clear and delete chains.

So this is wrapping iptables calls at the end of the day.
I don't see anything wrong in calling other system binaries. This has performance impact but for us it's neglectable.

PR #70, opened 2018-04-30
remove permissive binding

fgimenez:
During the transition of our components to RBAC we needed to configure the test cluster with `permissive-binding`, so that the components that were not yet migrated would keep working in the tests. The migration finished a while ago, and now this permissive setting is masking some errors, for instance, chart-operator bootstrap e2e test in cluster-operator should be failing until https://github.com/giantswarm/cluster-operator/pull/93 lands, and it is passing https://circleci.com/gh/giantswarm/cluster-operator/439

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -126,9 +126,6 @@ func (p *Project) CommonTearDownSteps() error {
 			Run: "helm reset --force",
 		},
 		Step{
-			Run: "kubectl delete clusterrolebinding permissive-binding",
-		},
-		Step{
 			Run: "kubectl -n kube-system delete sa tiller",
 		},
 		Step{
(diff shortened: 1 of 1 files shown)


indyfree (commented):
Is this eventually going to break some tests when vendored, or are we sure the correct rbac is in place everywhere? I guess this is one way to find out ;)

kopiczko (approved):
Please let's also create an umbrella issue to update all operators running e2e tests.

fgimenez:
Sure here it is https://github.com/giantswarm/giantswarm/issues/3104 i think all the projects with e2e test are there, please update if not.

fgimenez:
@kopiczko in fact it's much easier, the setup is done by the binary, as soon as it is released there's no need to change anything in the consumer repos, they always download the latest binary release in each build.

PR #74, opened 2018-05-11
Add g8s client to guest package

rossf7:
Adds the g8sclient to the guest package. For use in cluster-operator tests for chartconfig resources.


fgimenez (approved):
<3

PR #75, opened 2018-05-31
Query versions from release index

fgimenez:
Changes to get versions to test from https://github.com/giantswarm/installations/tree/master/release/provider


rossf7 (approved):
LGTM. Very nice, I like that we can keep the same current and WIP functionality.

Thanks for taking care of this <3

PR #78, opened 2018-06-04
Sort index releases by semver

fgimenez:
Aims to get the right versions for e2e `WIP` and `current` builds, sorting by date doesn't get the more recent versions of each kind.


pkg/framework/version_bundle_test.go~
```diff
@@ -0,0 +1,7 @@
+package framework
+
+import "testing"
+
+func Test_SortBySemver(t *testing.T) {
+
+}
```
rossf7: Looks like this is a temp file.

pkg/framework/version_bundle.go
```diff
@@ -6,6 +6,7 @@ import (
 	"log"
 	"sort"
 
+	"github.com/blang/semver"
```
rossf7: I like the use of the semver library here.

rossf7 (approved):
LGTM apart from one temp file issue.

pkg/framework/version_bundle_test.go~
fgimenez: Very good catch thx! my global git ignore got lost, fixed

pkg/framework/version_bundle.go
tuommaki: I'm kind of late here, but is there something specific in that library compared to `github.com/coreos/go-semver/semver` used in `versionbundle`? FYI: There's also same implementation in `versionbundle` library: https://github.com/giantswarm/versionbundle/blob/master/releases_sort.go#L9 :)

fgimenez: Absolutely, I can take care of that, already put the changes here to take advantage of that.

fgimenez:
Updated to use the new IndexRelease sorter from versionbundle giantswarm/versionbundle#52 thx @tuommaki! Merging.

PR #80, opened 2018-06-08
Add GetAuthorities method

fgimenez:
This will allow feeding versionBundles info into clusterconfig chart.

### DIFF
--- a/pkg/framework/version_bundle.go
+++ b/pkg/framework/version_bundle.go
@@ -46,6 +46,25 @@ func GetVersionBundleVersion(params *VBVParams) (string, error) {
 	return output, nil
 }
 
+func GetAuthorities(params *VBVParams) ([]versionbundle.Authority, error) {
+	err := checkType(params.VType)
+	if err != nil {
+		return []versionbundle.Authority{}, microerror.Mask(err)
+	}
+	log.Printf("Tested version %q", params.VType)
+
+	content, err := getContent(params.Provider, params.Token)
+	if err != nil {
+		return []versionbundle.Authority{}, microerror.Mask(err)
+	}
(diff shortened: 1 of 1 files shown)


pkg/framework/version_bundle.go
```diff
@@ -46,6 +46,25 @@ func GetVersionBundleVersion(params *VBVParams) (string, error) {
 	return output, nil
 }
 
+func GetAuthorities(params *VBVParams) ([]versionbundle.Authority, error) {
+	err := checkType(params.VType)
+	if err != nil {
+		return []versionbundle.Authority{}, microerror.Mask(err)
```
xh3b4sd: Zero value of a slice is `nil`.

pkg/framework/version_bundle.go
```diff
@@ -46,6 +46,25 @@ func GetVersionBundleVersion(params *VBVParams) (string, error) {
...
+	}
+	log.Printf("Tested version %q", params.VType)
```
xh3b4sd: I would like to use our logger as we do everywhere. This here will not play in favour of our perception when we humans deal with logs.

tuommaki (approved):
As Tim suggested, I'd also prefer `nil` instead of empty slice when returning error. Otherwise LGTM.

pkg/framework/version_bundle.go
fgimenez: Sure, done.

pkg/framework/version_bundle.go
fgimenez: Done, PTAL.

xh3b4sd (approved):

PR #81, opened 2018-06-08
Pass filesystem labels to docker volume mount

tuommaki:
When running e.g. on SELinux enabled system, filesystem labels need to
be passed to docker container in order to permissions work correctly.

These are needed e.g. when running on Fedora Linux with defaults.


fgimenez (approved):
LGTM

PR #83, opened 2018-06-15
some refactorings

xh3b4sd:


pkg/framework/backoff.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 const (
-	LongMaxWait  = 30 * time.Minute
+	LongMaxWait  = 40 * time.Minute
```
xh3b4sd: This is for Azure because it takes super long sometimes.

pkg/framework/backoff.go
```diff
@@ -30,3 +30,42 @@ func NewExponentialBackoff(maxWait, maxInterval time.Duration) *backoff.Exponent
 
 	return b
 }
+
+func NewConstantBackoff(maxWait, maxInterval time.Duration) backoff.BackOff {
```
xh3b4sd: This is to have less spam in the e2e tests. We can straight up retry after a constant 60 seconds or so when we wait for the API to get up. It does not make sense to print lots of messages for each retry after 1 second.

pkg/framework/backoff.go
```diff
@@ -30,3 +30,42 @@ func NewExponentialBackoff(maxWait, maxInterval time.Duration) *backoff.Exponent
...
+	b := WithMaxElapsedTime(backoff.NewConstantBackOff(maxInterval), maxWait)
+
+	b.Reset()
+
+	return b
+}
+
+func WithMaxElapsedTime(b backoff.BackOff, d time.Duration) *BackOffMaxElapsedTime {
```
xh3b4sd:
```go
package main

import (
	"fmt"
	"log"
	"time"

	"github.com/cenkalti/backoff"
)

func main() {
	o := func() error {
		return fmt.Errorf("test error")
	}
	b := NewConstantBackoff(ShortMaxWait, ShortMaxInterval)
	n := func(err error, delay time.Duration) {
		log.Println("level", "debug", "message", err.Error())
	}

	err := backoff.RetryNotify(o, b, n)
	if err != nil {
		panic(err)
	}
}

const (
	ShortMaxWait = 30 * time.Second
	ShortMaxInterval = 5 * time.Second
)

func NewConstantBackoff(maxWait, maxInterval time.Duration) backoff.BackOff {
	b := WithMaxElapsedTime(backoff.NewConstantBackOff(maxInterval), maxWait)

	b.Reset()

	return b
}

func WithMaxElapsedTime(b backoff.BackOff, d time.Duration) backoff.BackOff {
	return &BackOffElapsedTime{
		delegate:   b,
		maxElapsed: d,
		start:      time.Time{},
	}
}

type BackOffElapsedTime struct {
	delegate   backoff.BackOff
	maxElapsed time.Duration
	start      time.Time
}

func (b *BackOffElapsedTime) NextBackOff() time.Duration {
	if b.start.IsZero() {
		b.start = time.Now()
	}

	if time.Now().After(b.start.Add(b.maxElapsed)) {
		return backoff.Stop
	}

	return b.delegate.NextBackOff()
}

func (b *BackOffElapsedTime) Reset() {
	b.start = time.Time{}
	b.delegate.Reset()
}
```

```
$ go run main.go
2018/06/15 15:56:11 level debug message test error
2018/06/15 15:56:16 level debug message test error
2018/06/15 15:56:21 level debug message test error
2018/06/15 15:56:26 level debug message test error
2018/06/15 15:56:31 level debug message test error
2018/06/15 15:56:36 level debug message test error
panic: test error

goroutine 1 [running]:
main.main()
	/Users/xh3b4sd/go/src/github.com/giantswarm/test/main.go:22 +0xa0
exit status 2
```

pkg/framework/host.go
```diff
@@ -102,6 +102,47 @@ func (h *Host) AWSCluster(name string) (*v1alpha1.AWSConfig, error) {
 	return cluster, nil
 }
 
+func (h *Host) CreateNamespace(ns string) error {
```
xh3b4sd: In this file all changes are only reorderings.

pkg/framework/version_bundle.go
```diff
@@ -48,7 +47,9 @@ func GetVersionBundleVersion(params *VBVParams) (string, error) {
 		return "", microerror.Mask(err)
 	}
 
-	logger.Log("level", "debug", fmt.Sprintf("Version Bundle Version %q", output))
+	logger.Log("level", "debug", "message", fmt.Sprintf("tested version '%s'", params.VType))
+	logger.Log("level", "debug", "message", fmt.Sprintf("version bundle version '%s'", output))
```
xh3b4sd: Here are only some log improvements.

pkg/framework/version_bundle.go
```diff
@@ -107,7 +109,7 @@ func checkType(vType string) error {
 		}
 	}
 	if !isValid {
-		return fmt.Errorf("%q is not a valid version bundle version type", vType)
+		return fmt.Errorf("'%s' is not a valid version bundle version type", vType)
```
rossf7: Nitpick but I think this is fine with the `%q`.  We use it in lots of places and I find it easy to read.

pkg/framework/backoff.go
rossf7: This is sad but necessary I guess.

pkg/framework/backoff.go
rossf7: Nice I think this is a better interval and less noisy.

rossf7 (approved):
LGTM just one nitpick on the logging changes.

pkg/framework/version_bundle.go
xh3b4sd: With my change the first line is more readable IMO than the double quoted one, because this is the rendered result we really see. 

```
"'%s' is not a valid version bundle version type"
"\"%s\" is not a valid version bundle version type"
```

rossf7: OK fine thanks for the explanation.

PR #85, opened 2018-07-10
Allow CLUSTER_NAME and COMMON_DOMAIN to be given in GuestConfig

tuommaki:
There are use cases where multiple guest clusters need to be created
concurrently and use of environment variables for CLUSTER_NAME is
slightly difficult to safely manage. Therefore allow it to be specified
in GuestConfig. To align with this, also COMMON_DOMAIN can be defined
via config variable. When either of given fields is empty, this
implementation defaults to environment variable in order to maintain
current functionality.

### DIFF
--- a/pkg/framework/guest.go
+++ b/pkg/framework/guest.go
@@ -1,6 +1,7 @@
 package framework
 
 import (
+	"fmt"
 	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


fgimenez (approved):
LGTM

xh3b4sd:
IMO we should constantly work towards removing the environment support from any library. The actual tools being executed should be the only instances working with the environment, if necessary. From there configuration should be injected down the road as usual.

xh3b4sd:
DO we actually agree on this?

tuommaki:
> Do we actually agree on this?

What do you mean by `we`? I do agree, but AFAIU this is a bit of work to do and I see it best to progress in similar way that I did in this PR. Otherwise it's again a long time out-of-teamwork when fixing the whole suite + tests in all providers and verify running them correctly etc.

PR #88, opened 2018-07-17
env refactoring

xh3b4sd:


pkg/framework/guest.go
```diff
@@ -25,42 +24,44 @@ const (
 )
 
 type GuestConfig struct {
-	ClusterName  string
```
xh3b4sd: The cluster ID is a setting, not a dependency. Moving.

pkg/framework/guest.go
```diff
@@ -25,42 +24,44 @@ const (
...
 	restConfig *rest.Config
+
+	clusterID    string
+	commonDomain string
 }
 
 func NewGuest(config GuestConfig) (*Guest, error) {
-	if config.ClusterName == "" {
-		// If config.ClusterName is not defined, default to environment variable.
-		config.ClusterName = os.Getenv("CLUSTER_NAME")
```
xh3b4sd: Env magic should only ever be done on program initialization. Not somewhere in the middle. Having it configurable is way easier to maintain. Also the compiler can do its part and help solving problems with the code. Also the responsibility of who has to wire it is way more clear.

pkg/framework/guest.go
```diff
@@ -105,14 +106,14 @@ func (g *Guest) Initialize() error {
 	var guestK8sClient kubernetes.Interface
 	var guestRestConfig *rest.Config
 	{
-		n := fmt.Sprintf("%s-api", g.clusterName)
+		n := fmt.Sprintf("%s-api", g.clusterID)
```
xh3b4sd: Our terminology is cluster ID. It ever was.

fgimenez (approved):
LGTM

pkg/framework/host.go
```diff
@@ -48,11 +52,14 @@ func NewHost(c HostConfig) (*Host, error) {
 		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", c)
 	}
 
+	if c.ClusterID == "" {
```
MarcelMue: A little late on this one, but E2E tests without a Guest cluster don't need this AFAIK.
We should default to something here or separate host from guest cluster magic structurally.

xh3b4sd: The host and guest frameworks are already separated. This started out as being used against the provider operators where we always have guest clusters we test against. When there is functionality we need in e2e tests where no guest cluster exists, we should extract this functionality to make it reusable.

rossf7: We need to be able to run e2e tests without a guest cluster. As well as chart-operator itself the Ingress Controller chart also has e2e tests. 

https://github.com/giantswarm/kubernetes-nginx-ingress-controller

These are now broken after vendoring the latest e2e-harness. To be fair they hadn't been updated in a while. I'll see if I can extract the logic we need as part of fixing them.

xh3b4sd: I see. Can I help sorting this somehow?

rossf7: Not sure yet.

The problem is the host framework has a lot of generic and useful functionality like creating namespaces or installing charts. These is also a lot of guest cluster specific logic like deleting guest cluster CRs and installing vault.

I think it would be good to split the generic logic out. WDYT?

MarcelMue: I am working on this, also read https://gigantic.slack.com/archives/C3C7ZQXC1/p1531987445000221

I will first abstract out resource related functionality and add in `helmClient` and `apprClient`.

xh3b4sd: Would probably be good to split what is not guest cluster related. The chart installation magic is also using the shell out stuff which should definitely rather use the helm client lib to have proper error handling and so on.

MarcelMue: Defaulting here could work in the meantime, but I don't think anyone is going to update the lib until I am done anyway IMO.

rossf7: Marcel, thanks for posting that thread. Yes that's exactly this issue. Let me know if I can help with that?

Tim, yes good to update to use helmclient where we can. I'm happy to help with that tidying up.

The Ingress Controller e2e tests are out of date with our current structure. e.g. no setup directory. As a first step I'll update to the latest structure and e2e-harness and default the Cluster ID to get them passing. 

More refactoring can then be done in later steps. Thx for the info here!

PR #91, opened 2018-07-19
Introduce client libs and implement new InstallResource

MarcelMue:


Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN apk -Uuv add --update --no-cache \
       less=520-r0 \
       libffi-dev=3.2.1-r4 \
       openssh-client=7.5_p1-r8 \
-      openssl=1.0.2o-r0 \
+      openssl=1.0.2o-r1 \
```
MarcelMue: Needed to update this to make the build pass.

Gopkg.toml
```diff
@@ -24,6 +24,15 @@
 #   go-tests = true
 #   unused-packages = true
 
+required = [
+ "k8s.io/apiserver/pkg/endpoints/request"
+]
+
+# Fix:
+#   vendor/k8s.io/kubernetes/pkg/util/parsers/parsers.go:36:16: undefined: reference.ParseNormalizedNamed
```
MarcelMue: Found this workaround in a istio repo, any better idea how I can fix this?

Gopkg.toml
```diff
@@ -24,6 +24,15 @@
 #   go-tests = true
 #   unused-packages = true
 
+required = [
+ "k8s.io/apiserver/pkg/endpoints/request"
```
MarcelMue: Workaround: Added to prevent build errors in k8s.

pkg/framework/wait.go
```diff
@@ -11,7 +11,7 @@ const (
 	defaultTimeout = 1800
 )
 
-func waitFor(f func() error) error {
+func WaitFor(f func() error) error {
```
MarcelMue: Made this public to reuse it in the new package.

pkg/framework/resource/resource.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	return c, nil
+}
+
+func (r *Resource) InstallResource(name, values, version string, conditions ...func() error) error {
+	chartValuesEnv := os.ExpandEnv(values)
+	chartname := fmt.Sprintf("%s-chart", name)
+
+	tarball, err := r.apprClient.PullChartTarballFromRelease(chartname, version)
```
MarcelMue: Using `PullChartTarballFromRelease` here because we want to keep the functionality of `InstallBranchOperator` later.

pkg/framework/resource/resource.go
```diff
@@ -0,0 +1,68 @@
...
+	c := &Resource{
+		apprClient: config.ApprClient,
+		helmClient: config.HelmClient,
+		logger:     config.Logger,
+	}
+
+	return c, nil
+}
+
+func (r *Resource) InstallResource(name, values, version string, conditions ...func() error) error {
```
MarcelMue: The old function still exists, there are a couple of other things to do before we can remove it. I want to get this in as a first step and to have the updated vendor in.

Gopkg.toml
xh3b4sd: We do this in other places as well. These errors occur in pairs as it seems.

pkg/framework/wait.go
xh3b4sd: I am for dropping this and just using the backoff thing. Like why do we have to implement this again and use a different thing that does the same? It is harder to configure and when checking the logs of e2e tests its behaviour is then different. Don't understand that.

pkg/framework/resource/resource.go
```diff
@@ -0,0 +1,68 @@
...
+	chartValuesEnv := os.ExpandEnv(values)
+	chartname := fmt.Sprintf("%s-chart", name)
+
+	tarball, err := r.apprClient.PullChartTarballFromRelease(chartname, version)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	err = r.helmClient.InstallFromTarball(tarball, "giantswarm", helm.ReleaseName(name), helm.ValueOverrides([]byte(chartValuesEnv)), helm.InstallWait(true))
```
xh3b4sd: The org or whatever that is should be configurable. If this is the default then let's default it but having it hard coded is not so cool.

xh3b4sd (dismissed):
This makes total sense but I would like to see the backoff being used.

fgimenez (approved):
LGTM <3

pkg/framework/wait.go
MarcelMue: Done, this will need cleaning as well - we have to `pkg/framework/wait.go` file, `pkg/framework/backoff.go`, `pkg/framework/retry.go` and `pkg/wait/wait.go`.

pkg/framework/resource/resource.go
MarcelMue: This is a weird one: It's a `namespace` - but I could not easily find what the namespace is used for. In the cli there is optional namespace value but this one seems required.

I wanted to hardcode it here and then test what it actually does or which namespace is meant. I couldn't find any documentation which explains if this is the namespace flag similar to the cli or a tiller namespace or something else.

Maybe @fgimenez  has an idea?

MarcelMue: In our operators we just set it to the namespace of the chart resource in the customObject we watch, doesn't make it much clearer to me.

fgimenez: The signature of `helmClient.InstallFromTarball` mimics upstream's `InstallRelease` https://github.com/helm/helm/blob/master/pkg/helm/client.go#L78 in fact it is mostly a proxy method that adds a retry wrapper. Upstream's `InstallRelease` requires a namespace, so does our `helmClient.InstallFromTarball`. 

Regarding the `giantswarm` value, in our current e2e tests all components are installed in this namespace, the same as they are on production. I think it makes sense to parameterize this value and define the default value as a constant. 

Let me know if this makes things more clear?

MarcelMue: Thanks, I was aware of the proxy to upstream - I kind of don't get the point of the namespace upstream. Charts we install have namespaces set on the individual resources - why do we declare this here again? And it isn't mandatory in helm cli IIRC.

I will add the default like you said, still a tiny bit of confusion.

MarcelMue: Namespace is in the config now and gets defaulted to `giantswarm`.

MarcelMue:
Going with this here, I addressed review comments. Feel free to ping me if you want to see some changes.

PR #92, opened 2018-07-20
Adds vaultToken param

fgimenez:
It doesn't fail on Host instantiation because not all the tests require a Vault instance.


MarcelMue (approved):
Change LGTM, build failed but merge master and it should be fine.

kopiczko (approved):

xh3b4sd:
All e2e tests fail now with an updated `e2e-harness` lib because the Vault token magic is not done in e.g. `azure-operator` and `aws-operator`.

pkg/framework/host.go
```diff
@@ -413,13 +416,17 @@ func (h *Host) crd(crdName string) func() error {
 }
 
 func (h *Host) installVault() error {
+	if h.vaultToken == "" {
+		return microerror.Mask(missingVaultTokenError)
+	}
```
xh3b4sd: IMO this should be done in the constructor. See https://github.com/giantswarm/e2e-harness/pull/95.

PR #93, opened 2018-07-20
Default the appr client

MarcelMue:
Just an idea I just had: We will most likely use the same config for the appr client in all our e2e tests, so we could just default it most of the time. WDYT?

### DIFF
--- a/pkg/framework/resource/resource.go
+++ b/pkg/framework/resource/resource.go
@@ -10,6 +10,7 @@ import (
 	"github.com/giantswarm/helmclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
+	"github.com/spf13/afero"
 	"k8s.io/helm/pkg/helm"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/framework/resource/resource.go
```diff
@@ -34,7 +35,22 @@ func New(config ResourceConfig) (*Resource, error) {
 		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
 	}
 	if config.ApprClient == nil {
-		return nil, microerror.Maskf(invalidConfigError, "%T.ApprClient must not be empty", config)
+		config.Logger.Log("level", "warning", "message", fmt.Sprintf("%T.ApprClient is empty, using default", config))
```
xh3b4sd: I think this should just be debug. I also think when you need a comma in a message you should emit two logs.

xh3b4sd (approved):
I think this makes sense but I would tweak the logging a bit.

pkg/framework/resource/resource.go
MarcelMue: Changed to debug, do you mean two logs like this? Or did I misunderstand you

xh3b4sd: Yes I think this is better.

fgimenez (approved):
LGTM

pkg/framework/resource/resource.go
```diff
@@ -34,7 +35,24 @@ func New(config ResourceConfig) (*Resource, error) {
...
+		config.Logger.Log("level", "debug", "message", fmt.Sprintf("%T.ApprClient is empty", config))
+
+		config.Logger.Log("level", "debug", "message", fmt.Sprintf("using default for %T.ApprClient", config))
+
+		c := apprclient.Config{
+			Fs:     afero.NewOsFs(),
+			Logger: config.Logger,
+
+			Address:      "https://quay.io",
+			Organization: "giantswarm",
```
TheoBrigitte: Would be good to have those in the `ResourceConfig` or at least as constants.

TheoBrigitte (changes requested):
Just one request

pkg/framework/resource/resource.go
MarcelMue: The whole `apprClient` is part of the `ResourceConfig` - I just create one here with values I defaulted sensibly. Adding constants fields weird because I am just using them once so far. 

How can I improve this ?

TheoBrigitte: I think constants are also good in order no to hide values down in the code, just my opinion.

MarcelMue: I will add constants later when I add more functionality here. Going with this for now.

PR #94, opened 2018-07-20
Remove duplicate backoff function

MarcelMue:


pkg/framework/retry.go
```diff
@@ -1,35 +0,0 @@
...
-
-	"github.com/cenkalti/backoff"
-)
-
-const maxElapsedTime = 2 * time.Minute
-
-// newCustomExponentialBackoff returns pointer to a backoff.ExponentialBackOff,
-// initialized with custom values. At the moment, we only override the
-// maxElapsedTime.
-func newCustomExponentialBackoff() *backoff.ExponentialBackOff {
```
MarcelMue: This is the same function as `NewExponentialBackoff` with one less input parameter. We used this once and can simply pass the default in the second value.

pkg/framework/retry.go
```diff
@@ -1,35 +0,0 @@
...
-
-import (
-	"fmt"
-	"log"
-	"time"
-
-	"github.com/cenkalti/backoff"
-)
-
-const maxElapsedTime = 2 * time.Minute
```
MarcelMue: This `const` already exists as `ShortMaxWait` in `backoff.go`.

kopiczko (approved):

fgimenez (approved):
LGTM

PR #95, opened 2018-07-21
fixes error handling for Vault token configuration

xh3b4sd:
Follow up of https://github.com/giantswarm/e2e-harness/pull/92/files.

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -57,6 +57,9 @@ func NewHost(c HostConfig) (*Host, error) {
 	if c.ClusterID == "" {
 		return nil, microerror.Maskf(invalidConfigError, "%T.ClusterID must not be empty", c)
 	}
+	if c.VaultToken == "" {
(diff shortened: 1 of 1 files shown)


MarcelMue (approved):
Fine with me

fgimenez:
Well, there are tests that don't require Vault, if we check for the token on the constructor we will need to set it  even when it is not required.

MarcelMue:
Yeah it could be better to default here instead of error now that I think about it.
Empty string should be fine though, maybe just log a warning?

MarcelMue:
The way I understood @xh3b4sd comment in the other PR the vaultToken isn't set in all projects e.g. aws-operator. I don't know much about the vault setup - you guys can probably judge better what works here.

xh3b4sd:
We had a discussion with Marcel already that we have e2e tests for guest cluster testing and those without. I understand the host and guest framework as being used for guest cluster testing. e2e tests not testing guest clusters and thus not requiring Vault and not requiring the host and guest setup should rely on other libs. So when there is common ground to be abstracted we should redesign the whole thing but not add more hacks to the libs.

PR #96, opened 2018-07-23
Remove wait implementation in framework

MarcelMue:
This was only used in the host file - a normal backoff should do basically the same or am I missing something?


xh3b4sd (approved):
Fine with me.

fgimenez (approved):
LGTM

indyfree (approved):
LGTM

PR #99, opened 2018-07-26
Add WaitForStatus and WaitForVersion to resource package

rossf7:
I'm updating the Ingress Controller chart e2e tests (WIP PR https://github.com/giantswarm/kubernetes-nginx-ingress-controller/pull/26).

This includes using the `resource` package to install the chart. But I need to wait for the release to be installed. @MarcelMue you've already done this in chart-operator but I think its better to move here.

https://github.com/giantswarm/chart-operator/blob/4b7de9d1e73306f08e34a17b8504e03f1ac07126/integration/release/release.go#L15

I also extracted `WaitForVersion` for completeness.

### DIFF
--- a/pkg/framework/resource/resource.go
+++ b/pkg/framework/resource/resource.go
@@ -2,7 +2,9 @@ package resource
 
 import (
 	"fmt"
+	"log"
 	"os"
+	"time"
 
 	"github.com/cenkalti/backoff"
 	"github.com/giantswarm/apprclient"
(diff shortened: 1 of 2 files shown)


pkg/framework/resource/resource.go
```diff
@@ -93,3 +96,51 @@ func (r *Resource) InstallResource(name, values, channel string, conditions ...f
 
 	return nil
 }
+
+func (r *Resource) WaitForStatus(gsHelmClient *helmclient.Client, release string, status string) error {
```
MarcelMue: HelmClient is part of the resource struct. Why pass it in again?

pkg/framework/resource/resource.go
```diff
@@ -93,3 +96,51 @@ func (r *Resource) InstallResource(name, values, channel string, conditions ...f
...
+			return microerror.Mask(err)
+		}
+		if rc.Status != status {
+			return microerror.Maskf(releaseStatusNotMatchingError, "waiting for %q, current %q", status, rc.Status)
+		}
+		return nil
+	}
+
+	notify := func(err error, t time.Duration) {
+		log.Printf("getting release status %s: %v", t, err)
```
MarcelMue: Do we use this different logging for better readability? I always find it hard to tell where logs are coming from.

pkg/framework/resource/resource.go
```diff
@@ -93,3 +96,51 @@ func (r *Resource) InstallResource(name, values, channel string, conditions ...f
...
+
+	b := framework.NewExponentialBackoff(framework.ShortMaxWait, framework.LongMaxInterval)
+	err := backoff.RetryNotify(operation, b, notify)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	return nil
+}
+
+func (r *Resource) WaitForVersion(gsHelmClient *helmclient.Client, release string, version string) error {
```
MarcelMue: Same here with helmClient.

pkg/framework/resource/resource.go
xh3b4sd: Lets please use the configured logger and the usual log structure.

rossf7 (commented):
Damn yes I moved this in a bit of a hurry :( Should be good now. PTAL.

xh3b4sd (approved):

PR #100, opened 2018-07-26
Allow use existing cluster

calvix:
towards https://github.com/giantswarm/giantswarm/issues/2440 
for KVM we already have precreated cluster so we don't wanna minikube and we dont wanna setup cluster on aws ( at least for now, maybe in future we will spawn kvm host cluster)

for https://github.com/giantswarm/kvm-operator/pull/522

### DIFF
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -45,6 +45,10 @@ var (
 	setupCmdTestDir string
 	name            string
 	existingCluster bool
+	k8sApiUrl       string
+	k8sCert         string
+	k8sCertCA       string
+	k8sCertKey      string
 	remoteCluster   bool
 )
 
@@ -53,8 +57,12 @@ func init() {
 
 	SetupCmd.Flags().StringVar(&setupCmdTestDir, "test-dir", project.DefaultDirectory, "Name of the directory containing executable tests.")
 	SetupCmd.Flags().StringVar(&name, "name", "e2e-harness", "CI execution identifier")
-	SetupCmd.Flags().BoolVar(&remoteCluster, "remote", true, "use remote cluster")
 	SetupCmd.Flags().BoolVar(&existingCluster, "existing", false, "can be used with --remote=true to use already existing cluster")
+	SetupCmd.Flags().StringVar(&k8sApiUrl, "k8s-api-url", "", "k8s api url for existing cluster")
+	SetupCmd.Flags().StringVar(&k8sCert, "k8s-cert", "", "k8s cert for auth for existing cluster")
+	SetupCmd.Flags().StringVar(&k8sCertCA, "k8s-cert-ca", "", "k8s cert ca for auth for existing cluster")
+	SetupCmd.Flags().StringVar(&k8sCertKey, "k8s-cert-key", "", "k8s cert key for auth for existing cluster")
+	SetupCmd.Flags().BoolVar(&remoteCluster, "remote", true, "use remote cluster")
 }
 
 func runSetup(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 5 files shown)


pkg/cluster/cluster.go
```diff
@@ -16,25 +16,30 @@ import (
...
+		remoteCluster:   remoteCluster,
 	}
 }
 
 // Create is a Task that creates a remote cluster or, if we
 // are using a local one, puts in place the required files for
 // later access to it
 func (c *Cluster) Create() error {
+	if c.existingCluster {
+		return nil
```
calvix: maybe it make sense to create kubeconfig here instead of in kvm-operator e2e tests?

fgimenez: :+1: As discussed on slack this would involve more changes, can be addressed in separate PR(s) if needed.

cmd/setup.go
```diff
@@ -59,6 +76,27 @@ func runSetupError(cmd *cobra.Command, args []string) error {
 		return microerror.Mask(err)
 	}
 
+	k8sApiUrl := os.Getenv(EnvVarK8sApiUrl)
```
fgimenez: I don't think it's a good idea to use this env vars here, we are passing all the options as flags.

cmd/setup.go
```diff
@@ -95,10 +133,19 @@ func runSetupError(cmd *cobra.Command, args []string) error {
 	}
 	p := project.New(pDeps, pCfg)
 	hCfg := harness.Config{
-		RemoteCluster: remoteCluster,
+		ExistingCluster: existingCluster,
+		RemoteCluster:   remoteCluster,
 	}
 	h := harness.New(logger, fs, hCfg)
-	c := cluster.New(logger, fs, d, remoteCluster)
+	c := cluster.New(logger,
```
fgimenez: IMO we should use a config object here.

cmd/teardown.go
```diff
@@ -80,11 +80,19 @@ func runTeardown(cmd *cobra.Command, args []string) error {
 		Fs:     fs,
 	}
 	p := project.New(pDeps, pCfg)
-	c := cluster.New(logger, fs, d, cfg.RemoteCluster)
+	c := cluster.New(logger,
+		fs,
+		d,
+		cfg.ExistingCluster,
+		cfg.RemoteCluster,
+		"",
```
fgimenez: Indeed, this doesn't look fine without a config object.

pkg/cluster/cluster.go
```diff
@@ -20,21 +21,46 @@ type Cluster struct {
...
+		kubeconfigFilePath, err := getKubeConfigPath()
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		err = c.createKubeconfig(kubeconfigFilePath)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		c.logger.Log("info", "Created kubeconfig minikube assets accessible for the test container")
```
fgimenez: minikube is wrong here.

calvix:
@fgimenez Moved kubeconfig creation here, can you take one more look if this looks good to you? Thanks!

cmd/setup.go
```diff
@@ -59,6 +76,27 @@ func runSetupError(cmd *cobra.Command, args []string) error {
 		return microerror.Mask(err)
 	}
 
+	k8sApiUrl := os.Getenv(EnvVarK8sApiUrl)
+	k8sCert := os.Getenv(EnvVarK8sCert)
+	k8sCertCA := os.Getenv(EnvVarK8sCertCA)
+	k8sCertPrivate := os.Getenv(EnvVarK8sCertPrivate)
```
calvix: does it make sense to have this imported via flags or rather via envs?
from e2e it doesnt really matter but for human use we would have flags which have really long values like
```
--cert-tls=Lsfaseoiqwfioonrbvowrbnwrobndosvvsdvsvoewoneovewovneoivnweovnewoivnewiovniweovnweiovnewoivnewiovnewiovnweoivnewiovnweoivvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvveiovniowrnvdknroiejabfdvuigehjkfvdehuivjdsnkreuvsdhjredcjrvhdinjsehufvdsnefhvsdnklevhndkjscfevijhdksxfhvdnkjsevhnijdsaclehvnsdjksdvnjkacsdvnjamrvasmrhvdawrhbuaschwedascnmhjvnsdkcsdvnkjclfsdvnkjmfvsdnkjcmsdvkjncmlsdvkjncmsdvnkjcmskdvnjmsdvjnkcmasdvnkjcmdvsnjkaclsdvjnk
 ```
as its double base64 encoded data
```

fgimenez: IMO better flags, at the end we will see on the CI script something like `--cert-tls=${CERT_TLS}`.

calvix:
migrated to flags instead of envs

PR #101, opened 2018-07-27
use-custom-k8s-context

calvix:
By default, e2e-harness has hardcoded `minikube` k8s context so this little change allows use context from existing cluster template

towards e2e for kvm https://github.com/giantswarm/giantswarm/issues/2440

### DIFF
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -34,6 +34,7 @@ var (
 	k8sCert         string
 	k8sCertCA       string
 	k8sCertKey      string
+	k8sContext      string
 	remoteCluster   bool
 )
 
@@ -46,6 +47,7 @@ func init() {
 	SetupCmd.Flags().StringVar(&k8sApiUrl, "k8s-api-url", "", "k8s api url for existing cluster")
 	SetupCmd.Flags().StringVar(&k8sCert, "k8s-cert", "", "k8s cert for auth for existing cluster")
 	SetupCmd.Flags().StringVar(&k8sCertCA, "k8s-cert-ca", "", "k8s cert ca for auth for existing cluster")
+	SetupCmd.Flags().StringVar(&k8sCertCA, "k8s-context", "minikube", "k8s context to use")
 	SetupCmd.Flags().StringVar(&k8sCertKey, "k8s-cert-key", "", "k8s cert key for auth for existing cluster")
 	SetupCmd.Flags().BoolVar(&remoteCluster, "remote", true, "use remote cluster")
 }
@@ -110,10 +112,6 @@ func runSetupError(cmd *cobra.Command, args []string) error {
 	pa := patterns.New(logger)
 	w := wait.New(logger, d, pa)
 
-	k8sContext := "minikube"
-	if existingCluster {
-		k8sContext = "giantswarm-e2e"
-	}
 	pCfg := &project.Config{
 		Name:       projectName,
 		K8sContext: k8sContext,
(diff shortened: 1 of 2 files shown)


pkg/project/project.go
```diff
@@ -80,7 +81,7 @@ func (p *Project) CommonSetupSteps() error {
 	p.logger.Log("info", "executing common setup steps")
 	steps := []Step{
 		Step{
-			Run: "kubectl config use-context minikube",
+			Run: "kubectl config use-context " + p.cfg.K8sContext,
```
kopiczko: When we are here I would rather prefer to run all commands with `--context` flag instead.

cmd/setup.go
```diff
@@ -110,8 +110,12 @@ func runSetupError(cmd *cobra.Command, args []string) error {
...
 	pCfg := &project.Config{
-		Name: projectName,
-		Tag:  projectTag,
+		Name:       projectName,
+		K8sContext: "minikube",
+		Tag:        projectTag,
+	}
+	if existingCluster {
+		pCfg.K8sContext = "giantswarm-e2e"
 	}
```
kopiczko: What does `existingCluster` mean? When is it set?

calvix: its flag, its set when running `e2e-harness setup` command.
Its configuration where e2e is using remote cluster but it won't create it as it will use an existing cluster.

pkg/project/project.go
calvix: I cannot say if it is used also somewhere else or if some other part is relying upon

cmd/setup.go
kopiczko: BTW to avoid confusion I think it's better to determine value before initializing the struct:

```
k8sContext := "minikube"
if existingCluster {
	k8sContext = "giantswarm-e2e"
}
pCfg := &project.Config{
	Name: projectName,
	Tag:  projectTag,
	Name:       projectName,
	K8sContext: k8sContext,
	Tag:        projectTag,
}
```

calvix: That would probably complicate things as when running teardown steps it already expects that correct context is used. Lots of stuff would have to be rewritten. Maybe that is out of scope for this PR?

kopiczko: Staring new discussion: I think people rely on that when doing testing on local machine. I think it would be good to add a flag like `e2e-harness setup --kubecontext ...` and it should default to `minikube`. Then it's fully backward compatible.

kopiczko: But before doing effort here I'd wait for others opinions. I'm far from being an expert here.

pkg/project/project.go
kopiczko: I think this comment https://github.com/giantswarm/e2e-harness/pull/101#discussion_r205736077 belongs here. Yes this is out of scope and definitely not a blocker here.

cmd/setup.go
calvix: fixed

calvix: I was thinking about it too, but the thing is that the name of context is decided inside of the code.
Its either `minikube` when minikube is spawned or `giantswarm-e2e` when existing cluster is used ( there is template frot he kubeconfig in the e2e-harness code) so user from outside doesnt have any real option ( unless he change code)

calvix:
anything else?

cmd/setup.go
kopiczko: > thing is that the name of context is decided inside of the code

Yes it was just hardcoded. But when you make it a flag and default to `minikube` this is backward compatible, and doesn't mess local testing.

pkg/project/project.go
xh3b4sd: I would also rather prefer adding a flag to the `e2e-harness` command to accept the context to use. This is way more flexible and transparent, instead of some magic we hide internally and everybody just has to know.

calvix:
flag for k8sContext added

PR #102, opened 2018-07-30
Add update function to resource

MarcelMue:
Adds a function to update a resource now using our clients.

### DIFF
--- a/pkg/framework/resource/resource.go
+++ b/pkg/framework/resource/resource.go
@@ -100,7 +100,7 @@ func (r *Resource) UpdateResource(name, values, channel string, conditions ...fu
 	chartValuesEnv := os.ExpandEnv(values)
 	chartname := fmt.Sprintf("%s-chart", name)
 
-	tarballPath, err := r.apprClient.PullChartTarball(chartname, "stable")
+	tarballPath, err := r.apprClient.PullChartTarball(chartname, channel)
 	if err != nil {
 		return microerror.Mask(err)
 	}
--- a/pkg/framework/resource/resource.go
+++ b/pkg/framework/resource/resource.go
@@ -2,7 +2,6 @@ package resource
 
 import (
 	"fmt"
-	"os"
 	"time"
 
 	"github.com/cenkalti/backoff"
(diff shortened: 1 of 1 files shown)


pkg/framework/resource/resource.go
```diff
@@ -96,6 +96,23 @@ func (r *Resource) InstallResource(name, values, channel string, conditions ...f
...
+func (r *Resource) UpdateResource(name, values, channel string, conditions ...func() error) error {
+	chartValuesEnv := os.ExpandEnv(values)
+	chartname := fmt.Sprintf("%s-chart", name)
+
+	tarballPath, err := r.apprClient.PullChartTarball(chartname, channel)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = r.helmClient.UpdateReleaseFromTarball(name, tarballPath, helm.UpdateValueOverrides([]byte(chartValuesEnv)))
```
indyfree: What happens if `chartValuesEnv` is empty? Will this still work?

MarcelMue: Yes, IIRC we actually have to pass _at least_ an empty string in here. If the string is empty, then the defaults inside the chart will be used.

pkg/framework/resource/resource.go
```diff
@@ -96,6 +96,23 @@ func (r *Resource) InstallResource(name, values, channel string, conditions ...f
 	return nil
 }
 
+func (r *Resource) UpdateResource(name, values, channel string, conditions ...func() error) error {
+	chartValuesEnv := os.ExpandEnv(values)
```
xh3b4sd: I am totally against all kinds of env magic here. Again. This is a library. In its depth no env magic should take place. The primitives should work against deterministic inputs.

xh3b4sd (changes requested):

pkg/framework/resource/resource.go
MarcelMue: I agree, the reason I left this here is that most of our `values` we use currently are expanded through env. I am fine with removing it, but then we will need to have a more complex transition when we want to use these functions in most of our e2e tests.

xh3b4sd: Why? When it is necessary to still use the env then you can just do it before calling the lib. Then it is visible in the tests.

pkg/framework/resource/resource.go
```diff
@@ -74,14 +73,13 @@ func New(config ResourceConfig) (*Resource, error) {
 }
 
 func (r *Resource) InstallResource(name, values, channel string, conditions ...func() error) error {
-	chartValuesEnv := os.ExpandEnv(values)
```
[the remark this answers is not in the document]
MarcelMue: Yes, we shouldn't really do the env stuff here.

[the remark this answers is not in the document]
kopiczko: Ok I understand now: https://github.com/giantswarm/e2e-harness/pull/102#discussion_r206129362

Are you going to go trough all projects and make sure everything is still fine? Or we just didn't use it at all?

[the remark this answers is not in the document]
MarcelMue: This function is not used by many E2E tests currently. Only `chart-operator`, most tests still use this function: https://github.com/giantswarm/e2e-harness/blob/master/pkg/framework/host.go#L227

So this change won't break anything, but I plan to move more projects to use this implementation with our proper clients.

kopiczko (approved):
LGTM thanks @MarcelMue for the effort to make things straight

xh3b4sd (approved):

indyfree (approved):
LGTM.

PR #104, opened 2018-07-31
disable-port-forward-for-existing-kvm-cluster

calvix:
disable port forwarding for the `existing cluster` option that is used for KVM

### DIFF
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -101,9 +101,10 @@ func runSetupError(cmd *cobra.Command, args []string) error {
 		c := docker.Config{
 			Logger: logger,
 
-			Dir:           setupCmdTestDir,
-			ImageTag:      e2eHarnessTag,
-			RemoteCluster: remoteCluster,
+			Dir:             setupCmdTestDir,
+			ExistingCluster: existingCluster,
+			ImageTag:        e2eHarnessTag,
+			RemoteCluster:   remoteCluster,
 		}
 
 		d = docker.New(c)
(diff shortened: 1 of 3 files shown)


xh3b4sd:
Please ping SIG operator and/or SIG releng for these reviews. For my knowledge, why don't we need to forward a port when a cluster already exists? That connection does not compute for me somehow.

calvix:
The port forwarding is there because  shipyard is creating minikube in aws, and minikube is not accessible from outside, so port forwarding makes sure that the remote minikube is accessible from the Circle machine.

In existing cluster(gastropod)  we have public k8s endpoint so there is no need for port forwarding.

indyfree:
> shipyard is creating minikube in aws?

Are we still using shipyard? Though this was long gone.

xh3b4sd:
Maybe fede can say something about this here and if the intentions of functionalities align. This is a bit out of my horizon right now.

indyfree:
I am pretty sure we don't use shipyard anymore. See this Thread: https://gigantic.slack.com/archives/C3C7ZQXC1/p1523371330000110

> i think we can safely move to `remote=false` by default, and even remove all the remote stuff from e2e-harness

kopiczko:
AIUI for kvm we have to reanimate remove stuff.

calvix:
I dont know if its used or not but code is still there.

kopiczko (approved):
I don't think you should be blocked here. When we decide to remove remote option both `remoteCluster` and `existingCluster` settings can be just removed

PR #106, opened 2018-08-01
Add deployment package to framework

indyfree:
Add reusable function to check if deployment properties are correct.

### DIFF
--- a/pkg/framework/deployment/deployment.go
+++ b/pkg/framework/deployment/deployment.go
@@ -10,7 +10,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-type DeploymentConfig struct {
+type Config struct {
 	K8sClient kubernetes.Interface
 	Logger    micrologger.Logger
 }
@@ -20,7 +20,7 @@ type Deployment struct {
 	logger    micrologger.Logger
 }
 
-func New(config DeploymentConfig) (*Deployment, error) {
+func New(config Config) (*Deployment, error) {
 	if config.K8sClient == nil {
 		return nil, microerror.Maskf(invalidConfigError, "%T.K8sClient must not be empty", config)
 	}
--- a/pkg/framework/deployment/deployment.go
+++ b/pkg/framework/deployment/deployment.go
@@ -40,29 +40,29 @@ func New(config DeploymentConfig) (*Deployment, error) {
 func (d *Deployment) Check(name string, replicas int, expectedLabels, expectedMatchLabels map[string]string) error {
 	ds, err := d.k8sClient.Apps().Deployments(metav1.NamespaceSystem).Get(name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		return microerror.Newf("could not find deployment: '%s' %v", name, err)
+		return microerror.Maskf(notFoundError, "could not find deployment: '%s'", name)
 	} else if err != nil {
-		return microerror.Newf("unexpected error getting deployment: %v", err)
+		return microerror.Mask(err)
 	}
 
 	// Check deployment labels.
 	if !reflect.DeepEqual(expectedLabels, ds.ObjectMeta.Labels) {
-		return microerror.Newf("expected labels: %v got: %v", expectedLabels, ds.ObjectMeta.Labels)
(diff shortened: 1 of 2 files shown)


pkg/framework/deployment/deployment.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return d, nil
+}
+
+// Check ensures that key properties of the deployment are correct.
+func (d *Deployment) Check(name string, replicas int, expectedLabels, expectedMatchLabels map[string]string) error {
+	ds, err := d.k8sClient.Apps().Deployments(metav1.NamespaceSystem).Get(name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return microerror.Newf("could not find deployment: '%s' %v", name, err)
```
xh3b4sd: Eh, no. See https://github.com/giantswarm/fmt/blob/master/go/errors.md.

xh3b4sd (changes requested):

pkg/framework/deployment/deployment.go
indyfree: Ahh :sweat_smile: too fast.

pkg/framework/deployment/deployment.go
```diff
@@ -0,0 +1,69 @@
...
+	"reflect"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+type DeploymentConfig struct {
```
MarcelMue: Can be removed to `Config`.

pkg/framework/deployment/deployment.go
indyfree: The way to go here is to create a custom error matcher i.e. `IsNotFoundError` and wrap/match the apierror there, right? Similar to the promtheus example?

xh3b4sd: Maybe. Depends on what you want to do. When the error here has to be managed in a special way then we should have control over it and return our own proper type.

MarcelMue: You could just use `microerror.Maskf` IMO. I am kind of wondering why there is a special case for `NotFound` to begin with. We don't do anything different with it, right?

indyfree: Decided to go for an own `notFoundError` type, so consuming packages can match it more easily. Also created and masked errors for the other cases. Is that done right?

xh3b4sd: Do we want to act differently though? My question would be why not one `incorrectDeploymentError` to rule them all? Just thinking loud here.

pkg/framework/deployment/deployment.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return d, nil
+}
+
+// Check ensures that key properties of the deployment are correct.
+func (d *Deployment) Check(name string, replicas int, expectedLabels, expectedMatchLabels map[string]string) error {
+	ds, err := d.k8sClient.Apps().Deployments(metav1.NamespaceSystem).Get(name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return microerror.Maskf(notFoundError, "could not find deployment: '%s'", name)
```
kopiczko: This error is printed like: `not found: could not find deployment: '...'`. Let's not repeat this information.`not found: deployment: '...'` is enough.

TheoBrigitte (approved):
LGTM

pkg/framework/deployment/deployment.go
indyfree: Good catch, done.

pkg/framework/deployment/deployment.go
indyfree: You have a good point `incorrectDeploymentError` is good to caputure all incorrect properties, be it labels or replicas. Still IMO `notFoundError` is semantically something different than `incorrect`, and can also cause different actions.

xh3b4sd (approved):
Not blocking here.

MarcelMue (approved):
Fine with me :+1:

PR #109, opened 2018-08-07
Handle checking for deleted releases

rossf7:


pkg/framework/resource/resource.go
```diff
@@ -113,7 +113,10 @@ func (r *Resource) UpdateResource(name, values, channel string, conditions ...fu
 func (r *Resource) WaitForStatus(release string, status string) error {
 	operation := func() error {
 		rc, err := r.helmClient.GetReleaseContent(release)
-		if err != nil {
+		if helmclient.IsReleaseNotFound(err) && status == "DELETED" {
+			// Error is expected because we purge releases when deleting.
+			return nil
+		} else if err != nil {
```
rossf7: I need to check if a helm release is deleted in my k8s-migrator test. Since we purge releases on deletion they never reach a deleted state.

This returns nil if the error is correct and we're looking for a deleted status.

xh3b4sd (approved):

PR #110, opened 2018-08-08
dont-install-tiller-for-existing-cluster

calvix:
There is no need to install tiller each time for the already existing cluster.  
( parallel e2e test could remove tiller in the middle of another test so rather have it installed from the start and don't do it in e2e)


fgimenez (approved):
LGTM

PR #113, opened 2018-08-08
rollback-namespace

calvix:
rollback of  https://github.com/giantswarm/e2e-harness/pull/112


xh3b4sd:
I don't think we need this. Looks like it works out in the provider operators with target namespace `default`.

PR #115, opened 2018-08-10
helm-release-name-with-ns-prefix

calvix:
Helm needs a unique name for each component and in KVM we will have several of them so we prefix with a namespace.

This will change the name from `xx-operator` -> `default-xx-operator` for helm release name.

Also adding function to be able get TargetNamespace in outside (for tearing down commands in operators)

This should not be destructive changes, as the only thing that will be affected is tearing down ( as the name changes) but that's not a big issue as after e2e the minikube is thrown away anyway.

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -381,6 +381,10 @@ func (h *Host) Setup() error {
 	return nil
 }
 
+func (h *Host) TargetNamespace() string {
+	return h.targetNamespace
+}
+
 func (h *Host) Teardown() {
 	HelmCmd("delete vault --purge")
 	h.k8sClient.CoreV1().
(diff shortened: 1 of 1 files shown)


fgimenez (commented):
Problem I see is with local executions, where the teardown is actually executed and needed to be able to rerun on each feedback loop. We will need to rewrite all of the teardown calls to remove helm releases.

calvix:
RIght, that is a problem. I guess that have to be done than, I don't see any other way how to workaround this,

TheoBrigitte (approved):
LGTM

calvix:
@fgimenez is there a way how could i find all the repos that are using the e2e-harness lib and need to be adjusted?

fgimenez:
@calvix i don't know of any way of finding them, I think currently they are `cert-operator`, `cluster-operator`, `azure-operator`, `aws-operator` and `chart-operator`.

PR #118, opened 2018-08-21
Increase ShortMaxWait for more stability

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/4016

### DIFF
--- a/pkg/framework/backoff.go
+++ b/pkg/framework/backoff.go
@@ -6,7 +6,7 @@ import (
 
 const (
 	LongMaxWait  = 40 * time.Minute
-	ShortMaxWait = 2 * time.Minute
+	ShortMaxWait = 4 * time.Minute
 )
 
 const (


pkg/framework/backoff.go
```diff
@@ -6,7 +6,7 @@ import (
 
 const (
 	LongMaxWait  = 40 * time.Minute
-	ShortMaxWait = 2 * time.Minute
+	ShortMaxWait = 4 * time.Minute
```
MarcelMue: Making this longer for chart-operator. Seems to be working - uncertain why we need more time now though.
See https://circleci.com/gh/giantswarm/chart-operator/tree/update-e2e-harness

indyfree (approved):
LGTM. Hopefully the flapping reduces with that. Would be nice to find out why we suddenly need more time though.

xh3b4sd:
Increasing timeouts and hoping for the problem to be magically fixed might not be what we want. Just saying.

MarcelMue:
Sure, this is to increase our quality of life right now as I don't want to invest too much time into this. I will keep the issue open to figure out what's actually wrong.

PR #120, opened 2018-08-30
Align default namespace

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/4029

We were not installing our operators in `giantswarm` and had different namespaces sprinkled around.
I think aligning to `giantswarm` makes sense.

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	defaultNamespace = "giantswarm"
+	defaultNamespace = "default"
 )
 
 type HostConfig struct {
@@ -464,7 +464,7 @@ func (h *Host) RestConfig() *rest.Config {
 }
 
 func (h *Host) Setup() error {
-	if err := h.CreateNamespace(h.targetNamespace); err != nil {
+	if err := h.CreateNamespace("giantswarm"); err != nil {
 		return microerror.Mask(err)
 	}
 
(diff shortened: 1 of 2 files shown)


xh3b4sd (changes requested):
I am pretty sure this will break everything and we had this already when Vaclav worked on KVM e2e. This is not that easy. We install CRs for instance in the default namespace for some reasons I do not recall. I mean, when you want to go with this make sure that all operator e2e tests work with it.

MarcelMue:
I see, then it was my mistake to introduce the `giantswarm` namespace stuff. I changed this PR to align to `default`. 

It's still a bit weird that we have everything in `default` in our e2e tests while we run our operators in `giantswarm` in prod. Something to consider in the future but not really critical.

kopiczko:
I agree when you want to mess with namespaces you should test with kvm first or talk with @calvix at least. Uncertain what implications it has now. But I appreciate you try to make those things straight.

MarcelMue:
This is just things I unknowingly changed to `giantswarm` before. This should not affect kvm e2e now I believe. Can you take another look?

xh3b4sd:
Lets hang and talk.

xh3b4sd (approved):
Not blocking here. Just think that the design is still that you can easily mess up everything with such changes so we have to try to get this straight in the future. When you think this is right, go for it.

PR #121, opened 2018-08-31
Make logger fallback to giantswarm namespace

MarcelMue:
We can't currently reliably tell which namespace an operator is going to be installed in. `e2e-harness` overwrites the `namespace` value in the helm values but not all operators actually use that value.

So we have some operators that end up in a namespace we pass in as a value while others do not. This attempts to fallback to the `giantswarm` namespace if we can't find the pod at first.

related issue: https://github.com/giantswarm/giantswarm/issues/4123

I will add a TODO issue to remove this code when the root cause is solved.

towards https://github.com/giantswarm/giantswarm/issues/4029

### DIFF
--- a/pkg/framework/host.go
+++ b/pkg/framework/host.go
@@ -324,8 +324,8 @@ func (h *Host) InstallOperator(name, cr, values, version string) error {
 	//
 	// Final version of the code:
 	//
-	// podName, err := h.PodName(h.targetNamespace, fmt.Sprintf("app=%s", name))
-	// 	if err != nil {
+	//	podName, err := h.PodName(h.targetNamespace, fmt.Sprintf("app=%s", name))
+	//	if err != nil {
 	//		return microerror.Mask(err)
 	//	}
 	//	err = h.filelogger.StartLoggingPod(h.targetNamespace, podName)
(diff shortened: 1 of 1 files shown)


pkg/framework/host.go
```diff
@@ -316,15 +316,41 @@ func (h *Host) InstallOperator(name, cr, values, version string) error {
...
+	//	if err != nil {
+	//		return microerror.Mask(err)
+	//	}
+	//
+	podNamespace := h.targetNamespace
+
+	podName, err := h.PodName(podNamespace, fmt.Sprintf("app=%s", name))
+	if IsNotFound(err) {
+		podNamespace = "giantswarm"
+		podName, err = h.PodName(podNamespace, fmt.Sprintf("app=%s", name))
```
xh3b4sd: To me it looks like we can directly use the target namespace everywhere but here and simply use the hardcoded alternative. The final version of the code would then be to remove the not found exception with its work around code of 4 lines.

xh3b4sd (approved):

pkg/framework/host.go
MarcelMue: We unfortunately also need it when we start the logging on the pod: https://github.com/giantswarm/e2e-harness/blob/ec5dbb5fb89a95b4b1d922ff13ee481c8cccdf2e/pkg/framework/host.go#L349

PR #125, opened 2018-09-17
Add delete resource function with helmclient

MarcelMue:
some grinding towards https://github.com/giantswarm/giantswarm/issues/3878

### DIFF
--- a/pkg/framework/resource/resource.go
+++ b/pkg/framework/resource/resource.go
@@ -76,7 +76,7 @@ func New(config Config) (*Resource, error) {
 	return c, nil
 }
 
-func (r *Resource) DeleteResource(name string) error {
+func (r *Resource) Delete(name string) error {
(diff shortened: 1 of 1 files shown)


pkg/framework/resource/resource.go
```diff
@@ -76,6 +75,15 @@ func New(config ResourceConfig) (*Resource, error) {
 	return c, nil
 }
 
+func (r *Resource) DeleteResource(name string) error {
```
xh3b4sd: Within the `resource` package on something that is created by `New` the method should rather be called `Delete`. Same with `Install` below.

xh3b4sd (approved):

pkg/framework/resource/resource.go
MarcelMue: 100% agreed, changed.

PR #129, opened 2018-09-19
Remove namespace fallback for file logger

MarcelMue:
https://github.com/giantswarm/giantswarm/issues/4123 got done and all namespaces of our operators are configurable now. We should no longer need to fall back to a static `giantswarm` namespace.


xh3b4sd (approved):

rossf7 (approved):
LGTM

PR #131, opened 2018-09-21
Use multiwrite to also write to stdout

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/4233

### DIFF
--- a/pkg/framework/filelogger/filelogger.go
+++ b/pkg/framework/filelogger/filelogger.go
@@ -90,8 +90,11 @@ func (f FileLogger) scan(readCloser io.ReadCloser, name string) {
 
 	defer outFile.Close()
 
+	var writers io.Writer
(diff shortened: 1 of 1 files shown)


MarcelMue:
@kopiczko please take a look if this would be helpful for debugging our issues in aws-operator.

MarcelMue:
Vendored into `aws-operator` here: https://circleci.com/gh/giantswarm/aws-operator/11670

MarcelMue:
Circle isn't playing nice here, uncertain if this has any benefit at all:
```
Your output is too large to display in the browser. Only the first 400000 characters are displayed.
You can download the output once this step is finished.
```

kopiczko:
> Circle isn't playing nice here, uncertain if this has any benefit at all

Yeah, totally forgot about that crap. I think it doesn't make much sense in that case.

MarcelMue:
Agreed, closing.

PR #134, opened 2018-10-01
Use time consts from backoff.

indyfree:
Towards https://github.com/giantswarm/giantswarm/issues/3878, https://github.com/giantswarm/giantswarm/issues/3896.

First small PR of a couple to get rid of the HelmCmd in `framework/host.go`.

When trying to reuse the `resource` package the conts introduced a circular dependency, so now taking them from the backoff lib.


xh3b4sd (approved):
Nice.

xh3b4sd:
This breaks other projects now. See for instance the builds of https://github.com/giantswarm/aws-operator/pull/1184. Please also see https://github.com/giantswarm/giantswarm/pull/4311. Please make sure to make the broken projects work now.

indyfree:
@xh3b4sd Sorry this happened. Was my shortsight. Seems like oponder provided a fix in https://github.com/giantswarm/e2etests/pull/17. I'll take care of vendoring the changes to the according projects.

xh3b4sd:
No worries. Happens to the best. My disappointment in the end came from the unfortunate discussions I had to have with people. We have to sort this out on a process level it seems since there are too many opinions right now.

PR #135, opened 2018-10-05
framework, vendor: fix waiting for CRD

kopiczko:
Fixes https://github.com/giantswarm/giantswarm/issues/4337.

- Vendor apiextensions-apiserver.
- Create apiextensions client.
- Use the client to poll for CRD existence.


pkg/framework/host.go
```diff
@@ -48,8 +49,9 @@ type Host struct {
 	logger     micrologger.Logger
 	filelogger *filelogger.FileLogger
 
+	extClient            *apiextensionsclient.Clientset
 	g8sClient            *versioned.Clientset
-	k8sClient            kubernetes.Interface
+	k8sClient            *kubernetes.Clientset
```
kopiczko: I just prefer to have most explicit type possible. That makes it easier to jump around the code and figure things. Also other clients where typed that way.

pkg/framework/host.go
```diff
@@ -565,12 +574,27 @@ func (h *Host) WaitForPodLog(namespace, needle, podName string) error {
...
+		}
+		b := backoff.NewExponential(backoff.ShortMaxWait, backoff.ShortMaxInterval)
+		n := backoff.NewNotifier(h.logger, ctx)
+		err := backoff.RetryNotify(o, b, n)
+
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		return nil
```
kopiczko: This is the most important part. It should fix builds like https://circleci.com/workflow-run/f929a4aa-b711-4902-933e-eedde9633670.

MarcelMue (commented):
I think the change is fine overall. Please create an issue on the releng board to vendor this into our projects. The build also currently fails, unsure why.

pkg/framework/host.go
```diff
@@ -310,16 +317,18 @@ func (h *Host) InstallStableOperator(name, cr, values string) error {
 	return nil
 }
 
-func (h *Host) InstallBranchOperator(name, cr, values string) error {
-	err := h.InstallOperator(name, cr, values, "@1.0.0-${CIRCLE_SHA1}")
+func (h *Host) InstallBranchOperator(name, crd, values string) error {
+	err := h.InstallOperator(name, crd, values, "@1.0.0-${CIRCLE_SHA1}")
```
indyfree: <3

pkg/framework/host.go
```diff
@@ -496,7 +505,7 @@ func (h *Host) PodName(namespace, labelSelector string) (string, error) {
 	return pod.Name, nil
 }
 
-// RestConfig returns the host cluster framework's rest config.
+// RestConfig returns a deep copy of host cluster framework's rest config.
```
indyfree: <3

pkg/framework/host.go
indyfree: Very nice!

indyfree (approved):
Cool!

pkg/framework/host.go
kopiczko: This is not actually true :) I tried to do the DeepCopy but it returns another type (uses embedded type DeepCopy). Good you commented though so I revert the comment :)

kopiczko:
Closing. This is implemented in `pkg/release` instead.

PR #139, opened 2018-10-09
pkg/release: revert "move from pkg/framework/resource"

kopiczko:
See comment https://github.com/giantswarm/giantswarm/issues/4355#issuecomment-427981428

TL;TR: the request is to prepare all vendor PRs upfront.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # e2e-harness
+add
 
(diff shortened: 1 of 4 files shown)


xh3b4sd (approved):

kopiczko:
CI isn't running for w/e reason. I don't want to block PPL and this is only revert so merging.

PR #142, opened 2018-10-09
reamde: split using project into categories

kopiczko:
This is the order in which we need to do vendor updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,16 +153,25 @@ the `DefaultKubeConfig` constant in the `giantswarm/e2e-harness/pkg/harness` pac
 
 ## Projects using `e2e-harness`
 
+### Test libraries
+
+- https://github.com/giantswarm/e2etests
+
+### Libraries
+
 - https://github.com/giantswarm/apprclient
+- https://github.com/giantswarm/certctl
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -171,7 +180,6 @@ the `DefaultKubeConfig` constant in the `giantswarm/e2e-harness/pkg/harness` pac
 - https://github.com/giantswarm/kubernetes-node-health
 - https://github.com/giantswarm/kvm-operator
 - https://github.com/giantswarm/net-exporter
-- https://github.com/giantswarm/operatorkit
 - https://github.com/giantswarm/release-operator
```
kopiczko: I just ran into issues where I need to do vendor updates in order. So I thought splitting projects into categories makes sense here.

xh3b4sd (approved):
Nice. Do you think it would make sense to write a sentence or two about the development process in the specific readmes?

MarcelMue (approved):
Thanks for the effort <3

kopiczko:
> Do you think it would make sense to write a sentence or two about the development process in the specific readmes?

I can't see how it could be not good, but let's maybe just paste link to some description in fmt to keep it DRY.

PR #144, opened 2018-10-10
pkg/release: add conditionSet

kopiczko:


pkg/release/condition_set.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"context"
+
+	"github.com/giantswarm/backoff"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
kopiczko: I don't like that we have to pull k8s.io deps to release package. But there is a helm client anyway, so I'm not very sad.

pkg/release/release.go
```diff
@@ -32,6 +35,8 @@ type Release struct {
 	logger     micrologger.Logger
 
 	namespace string
+
+	condition *conditionSet
```
kopiczko: Ready to use in InstallOperator.

pkg/release/condition_set.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+type conditionFn func() error
+
+type conditionSetConfig struct {
+	ExtClient *apiextensionsclient.Clientset
+	Logger    micrologger.Logger
+}
+
+type conditionSet struct {
```
kopiczko: `conditionSet` to not colide with `condition`.

pkg/release/condition_set.go
```diff
@@ -0,0 +1,62 @@
...
+		logger:    config.Logger,
+	}
+
+	return c, nil
+}
+
+func (c *conditionSet) CRD(ctx context.Context, name string) conditionFn {
+	return func() error {
+		o := func() error {
+			_, err := c.extClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(name, metav1.GetOptions{})
```
xh3b4sd: Do we have something similar in `operatorkit` already? Can we combine or refactor something there somehow? Just a quick brainwave.

pkg/release/release.go
```diff
@@ -22,6 +24,7 @@ type Config struct {
 	ApprClient *apprclient.Client
 	HelmClient *helmclient.Client
 	Logger     micrologger.Logger
+	RestConfig *rest.Config
```
xh3b4sd: IMO this is a setting and not a dependency.

pkg/release/condition_set.go
kopiczko: I don't think so. If you can point out exactly what it is we can maybe use it.
But I'm not so keen to pull operatorkit to testing libraries.

pkg/release/release.go
kopiczko: Done.

pkg/release/release.go
```diff
@@ -23,7 +25,8 @@ type Config struct {
 	HelmClient *helmclient.Client
 	Logger     micrologger.Logger
 
-	Namespace string
+	Namespace  string
+	RestConfig *rest.Config
```
MarcelMue: I am not sure if we want to pass the rest config in here, or pass the individual clients. How do we want to handle this? Now both `host` and this gets a rest config and creates its own clients.

MarcelMue: I would prefer just creating the clients in one place and passing them around tbh. Uncertain.

kopiczko: I was just worried about additional changes when we add more clients.
But maybe we should add all at once?

MarcelMue: You are right, let's pass the restconfig for now and do more changes later

kopiczko: All the clients added.

MarcelMue: Ack, should be fine. We'll see.

MarcelMue (approved):
LGTM

PR #145, opened 2018-10-10
Add InstallOperator fucntion

MarcelMue:


pkg/release/release.go
```diff
@@ -20,18 +23,22 @@ const (
 
 type Config struct {
 	ApprClient *apprclient.Client
+	FileLogger *filelogger.FileLogger
 	HelmClient *helmclient.Client
+	K8sClient  kubernetes.Interface
 	Logger     micrologger.Logger
 
-	Namespace string
+	TargetNamespace string
```
kopiczko: What's the reason behing calling it a target namespace? For me `Namespace` makes total sense.

MarcelMue: Because this is the target of our installation of resources, I can revert back to namespace if you prefer.

kopiczko: This is kind of obvious that this is the namespace where releases will be installed, isn't it?
I'd prefer to call it a `Namespace` yes, but no blocker.

MarcelMue: Reverted back to `namespace` PTAL.

kopiczko (approved):
LGTM.

kopiczko (changes requested):
https://github.com/giantswarm/e2e-harness/pull/145/files#r224035185

pkg/release/condition_set.go
```diff
@@ -48,7 +48,6 @@ func (c *conditionSet) CRD(ctx context.Context, name string) conditionFn {
 			} else {
 				return backoff.Permanent(microerror.Mask(err))
 			}
-			return nil
```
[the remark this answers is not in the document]
MarcelMue: Added back, the line was unreachable but the problem was the if statement in reality. Fixed it now.

pkg/release/condition_set.go
```diff
@@ -45,7 +45,7 @@ func (c *conditionSet) CRD(ctx context.Context, name string) conditionFn {
 			_, err := c.extClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(name, metav1.GetOptions{})
 			if apierrors.IsNotFound(err) {
 				return microerror.Mask(err)
-			} else {
+			} else if err != nil {
```
kopiczko: Good catch.

kopiczko (approved):

MarcelMue:
Going with this, let's iterate.

PR #146, opened 2018-10-10
pkg/release, vendor: add strong types & ctx

kopiczko:


pkg/release/version.go
```diff
@@ -0,0 +1,15 @@
+package release
+
+type Version string
```
kopiczko: This will allow `:` confusion and prevent us from adding additional function.

pkg/release/release.go
```diff
@@ -120,7 +121,7 @@ func New(config Config) (*Release, error) {
 	return r, nil
 }
 
-func (r *Release) Delete(name string) error {
+func (r *Release) Delete(ctx context.Context, name string) error {
```
kopiczko: I just added `ctx` to every external API call.

pkg/release/version.go
```diff
@@ -0,0 +1,15 @@
...
+
+func NewStableVersion() Version {
+	return Version(":stable")
```
MarcelMue: This won't be correct here, we do not need the `:` anymore when using helmclient.

pkg/release/version.go
```diff
@@ -0,0 +1,15 @@
...
+}
+
+func NewVersion(gitSHA string) Version {
+	return Version("@1.0.0-" + gitSHA)
```
MarcelMue: Uncertain if the @ is needed here.

kopiczko: Sho how `helmclient` differentiates between channels and tags?

kopiczko: PullChartTarballFromRelease

kopiczko: Fixed.

pkg/release/version.go
```diff
@@ -0,0 +1,24 @@
+package release
+
+type Version struct {
+	name      string
+	isChannel bool
+}
```
kopiczko: We can further extend it by adding `NewChannelVersion` and `NewReleaseVersion` but I don't want to expose more than necessary.

kopiczko (commented):
@MarcelMue PTALA

MarcelMue (approved):
We need to try all of this out and see how it works. Change seems fine.

PR #148, opened 2018-10-15
release: add and expose Secret install condition

kopiczko:
This will be useful for installing certificate secrets.

### DIFF
--- a/pkg/release/condition_set.go
+++ b/pkg/release/condition_set.go
@@ -10,17 +10,26 @@ import (
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
-type conditionFn func() error
+type ConditionFn func() error
+
+type ConditionSet interface {
+	// Secret return a function waiting for the Secret to appear in the
+	// Kubernetes API.
+	Secret(ctx context.Context, namespace, name string) ConditionFn
+}
 
 type conditionSetConfig struct {
 	ExtClient apiextensionsclient.Interface
+	K8sClient kubernetes.Interface
 	Logger    micrologger.Logger
 }
 
 type conditionSet struct {
 	extClient apiextensionsclient.Interface
+	k8sClient kubernetes.Interface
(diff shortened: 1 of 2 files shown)


pkg/release/condition_set.go
```diff
@@ -10,37 +10,50 @@ import (
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
-type conditionFn func() error
+type ConditionFn func() error
```
xh3b4sd: Same here with naming. I would like to go with `Func`.

pkg/release/release.go
```diff
@@ -121,6 +121,10 @@ func New(config Config) (*Release, error) {
 	return r, nil
 }
 
+func (r *Release) Condition() ConditionSet {
+	return r.condition
+}
```
xh3b4sd: I do not follow this here.

kopiczko: So we can do `release.Install(...., release.Condition().Secret(ctx, namespace, name))`.

xh3b4sd: I see. This makes sense. Just to be clear. `release` in your example is not the package but an instance of the service implementation, right?

kopiczko: Yeah.

pkg/release/condition_set.go
```diff
@@ -61,3 +74,25 @@ func (c *conditionSet) CRD(ctx context.Context, crd *apiextensionsv1beta1.Custom
 		return nil
 	}
 }
+
+func (c *conditionSet) Secret(ctx context.Context, namespace, name string) ConditionFn {
```
MarcelMue: The implication in naming so far was that this always means `secret exists` or `CRD exists` - I think this might be a little confusing for someone looking at the code currently. They will just see condition `secret`. Would it maybe make sense to rename functions to `SecretExists`? That way they would state a condition in the name.

MarcelMue (approved):
Make Tim happy and I am happy.

pkg/release/condition_set.go
kopiczko: I just based that on current CRD name.

kopiczko:
PTALA

xh3b4sd (approved):

PR #149, opened 2018-10-17
adds condition for having pods running

xh3b4sd:
We are going to use this for the new Vault chart installation.

### DIFF
--- a/pkg/release/condition_set.go
+++ b/pkg/release/condition_set.go
@@ -6,20 +6,13 @@ import (
 	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
+	"k8s.io/api/core/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
-
-type ConditionFunc func() error
-
-type ConditionSet interface {
-	// SecretExists return a function waiting for the Secret to appear in the
-	// Kubernetes API.
-	SecretExists(ctx context.Context, namespace, name string) ConditionFunc
-}
 
 type conditionSetConfig struct {
 	ExtClient apiextensionsclient.Interface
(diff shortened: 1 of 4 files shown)


pkg/release/spec.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "context"
+
+type ConditionFunc func() error
+
+type ConditionSet interface {
+	// SecretExists return a function waiting for the Secret to appear in the
+	// Kubernetes API.
+	SecretExists(ctx context.Context, namespace, name string) ConditionFunc
+}
```
kopiczko: I don't like moving that to spec.go. This isn't package spec. But fine.

`PodExists` is missing here.

pkg/release/condition_set.go
```diff
@@ -75,6 +68,35 @@ func (c *conditionSet) CRDExists(ctx context.Context, crd *apiextensionsv1beta1.
...
+				return microerror.Mask(err)
+			}
+			if len(pods.Items) > 1 {
+				return microerror.Mask(tooManyResultsError)
+			}
+
+			pod := pods.Items[0]
+			if pod.Status.Phase != v1.PodRunning {
+				return microerror.Maskf(unexpectedStatusPhaseError, string(pod.Status.Phase))
+			}
```
kopiczko: Nice.

pkg/release/release.go
```diff
@@ -187,6 +188,8 @@ func (r *Release) Install(ctx context.Context, name string, version Version, val
 		}
 	}
 
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensured creation of release %#q", name))
+
```
kopiczko: This code isn't idempotent so `ensure` doesn't fit there. It should be regular `creating`. I in general think we should avoid logging inside libraries, but let's see.

pkg/release/spec.go
kopiczko: I see you added it in the meantime. Please remove `CRDExists`. We don't use it outside and it wasn't exposed on purpose. With `CRDExists` this interface doesn't make sense as it hides nothing and isn't used for testing.

xh3b4sd: I do not follow why we should hide here anything. All the conditions are used for the test setup. Why hiding any of them? I do not follow the vision here.

pkg/release/release.go
xh3b4sd: Why isn't it idempotent though? I can fix the logs meanwhile but the method should maybe be improved in the future, shouldn't it?

pkg/release/spec.go
kopiczko: The api is wierd here. It requires `crd *apiextensionsv1beta1.CustomResourceDefinition`. We can decide to change thing w/o all vendor updates we'd need otherwise. That's why I'm extra careful with exposing anything.

pkg/release/release.go
kopiczko: I wouldn't say so. As we agreed in `fmt` recently idempotent version should be named `EnsureInstalled` so we know what to expect. I think it should provide better error handling rather than idempotency assuming we retain this name.

xh3b4sd: But we just want to install something and make sure what we want to be installed is installed so having an actual idempotent version to just ensure the e2e setup is correct would make sense.

pkg/release/spec.go
xh3b4sd: I see that we have `InstallOperator` which takes a `CRD` and does the condition magic internally. I do not see the reason to hide the condition. The CRD is provided as argument anyway so we could just accept the condition. Then we can maybe even get rid of `InstallOperator` and just use `Install` with the proper condition. So you would configure it with the CRD and pod condition for the operator and do the logging magic separate. This would be clean and not hide anything but be composable. This would then be done in `e2esetup` later anyway.

pkg/release/release.go
kopiczko: Not sure. We expect releases to not be installed in the control plane when we call those functions. I.e. we want a "fresh installation". Also those functions are called only once per release per test, so idempotency wouldn't bring much value there.

kopiczko: But still the function name doesn't reflect idempotency. The code isn't idempotent. When we make it idempotent changing two log messages is rather easy.

pkg/release/spec.go
kopiczko: `InstallOperator` does also some logging magic, if not that I would agree here to get rid of `InstallOperator`. Maybe at some point we decide to do it like that. I'm not so keen to remove it right now when we have so many changes already.

The reason to hide `CRDExists` it to make sure it's not used outside `e2e-harness` repo. This gives flexibility of changing internal code w/o worrying about vendoring. It's easy to expose it when needed. It might be hard to hide it again. I think this is a very basic programming principle.

xh3b4sd: Disagree and commit.

xh3b4sd: Removed `CRDExists` from the interface.

pkg/release/release.go
xh3b4sd: The log message is changed.

xh3b4sd:
Fishing for LGMT.

pkg/release/spec.go
kopiczko: No reason to be sad. We can add it any time. I disagreed and committed in this PR also. We are equal :D

kopiczko (approved):

PR #151, opened 2018-10-18
uses namespace to prefix Helm release

xh3b4sd:


pkg/release/release.go
```diff
@@ -163,6 +163,7 @@ func (r *Release) Install(ctx context.Context, name string, version Version, val
 	var err error
 
 	chartname := fmt.Sprintf("%s-chart", name)
+	releaseName := fmt.Sprintf("%s-%s", r.namespace, name)
```
kopiczko: Is that for kvm?

xh3b4sd: Yes, but this is still fishy as it seems.

MarcelMue: What was the problem you saw that you wanted to try this? 
This should be a problem because `release-names` should be unique but I never saw any issues with it weirdly.

There is also an upstream issue:
https://github.com/helm/helm/issues/2062

xh3b4sd: Well, with the host framework we did this already and we need it for KVM so I just replicate the behaviour we have right now but with the new release package.

MarcelMue (approved):
My brain was slow, let's do this.

xh3b4sd:
For the record I also adapted the release deletion to use the namespaced name.

PR #152, opened 2018-10-24
release: add some idempotency

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4481.

Next steps are to make filelogger idempotent. And start logging operator pods if release name has suffix `-operator`. That way we can get rid of `InstallOperator` long term. Also we'll have to expose `CRDExists` condition.

### DIFF
--- a/pkg/release/release.go
+++ b/pkg/release/release.go
@@ -132,9 +132,9 @@ func (r *Release) Delete(ctx context.Context, name string) error {
 
 	err := r.helmClient.DeleteRelease(releaseName, helm.DeletePurge(true))
 	if helmclient.IsReleaseNotFound(err) {
-		return microerror.Maskf(releaseNotFoundError, releaseName)
+		return microerror.Mask(err)
 	} else if helmclient.IsTillerNotFound(err) {
-		return microerror.Mask(tillerNotFoundError)
+		return microerror.Mask(err)
 	} else if err != nil {
 		return microerror.Mask(err)
 	}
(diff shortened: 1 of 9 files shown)


pkg/release/release.go
```diff
@@ -333,3 +373,48 @@ func (r *Release) podName(namespace, labelSelector string) (string, error) {
...
+
+	tarball, err := r.apprClient.PullChartTarballFromRelease(chart, version.String())
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	return tarball, nil
+}
+
+func (r *Release) waitForConditions(ctx context.Context, conditions ...func() error) error {
```
kopiczko: I think about exposing it and removing conditions from `Install` & `Update`. BTW `conditions` in `Update` aren't checked at all so maybe we don't even have to expose that ATM. Let's see.

pkg/release/release.go
```diff
@@ -128,78 +128,105 @@ func (r *Release) Condition() ConditionSet {
 func (r *Release) Delete(ctx context.Context, name string) error {
 	releaseName := fmt.Sprintf("%s-%s", r.namespace, name)
 
-	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting release %#q", releaseName))
```
kopiczko: I removed logs from `Install` & `Delete`. With error annotations those are only changes to those functions.

pkg/release/release.go
```diff
@@ -128,78 +128,105 @@ func (r *Release) Condition() ConditionSet {
...
-	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleted release %#q", releaseName))
-
 	return nil
 }
 
-func (r *Release) EnsureDeleted(ctx context.Context, name string) error {
-	releaseName := fmt.Sprintf("%s-%s", r.namespace, name)
+// EnsureDeleted makes sure the release is deleted and purged and all
+// conditions are met.
+func (r *Release) EnsureDeleted(ctx context.Context, name string, conditions ...func() error) error {
```
kopiczko: `EnsureDeleted` and `EnsureInstalled` are the new functions here. Diff is ugly, maybe look at the whole file.

pkg/release/release.go
```diff
@@ -128,78 +128,105 @@ func (r *Release) Condition() ConditionSet {
...
+// EnsureInstalled makes sure the release is installed and all conditions are
+// met.
+//
+// NOTE: It does not update the release if it already exists.
+func (r *Release) EnsureInstalled(ctx context.Context, name string, version Version, values string, conditions ...func() error) error {
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating release %#q", name))
 
-	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating release %#q", releaseName))
+		err := r.Install(ctx, name, version, values)
```
kopiczko: `conditions` aren't passed here on purpose. I added logging around them below. As stated in other comment I'd like to remove them soon.

pkg/release/release.go
xh3b4sd: Why though? I would like to see what is happening in the e2e tests. Does the Helm client do the logging now?

xh3b4sd (approved):

pkg/release/release.go
kopiczko: We should use `Ensure*` methods which do log. The problem I see that this methods return errors which may be handled and handling often ends up with just logging.

So then we have code like:

```
err := r.Delete(...)
if IsReleaseNotFound(err) {
log(err)
}
```

Then you have to know `Delete` does logging already, then you do some logging. Messages can change in this library and they are not aligned. Everything becomes a mess.

xh3b4sd: I talk about informational debug logging. You talk about error logging. These are different things. I think the informational debug logging should be kept and whoever wants to do error logging can do it and having both is just fine.

kopiczko: So let's take an example which can be `EnsureDeleted` method here. How should it look like? With those logs left we'd have output like:

```
deleting release `abc`
deleting release `abc`
release `abc` already deleted
```

When we have conventions we don't have to think we the method like this logs or not, because we know. Logging because good error handling is in place.

xh3b4sd: I do not follow where the logs come from and what you want to describe.

kopiczko: One `deleting release` logs from `EnsureDeleted` and second from `Delete` which is called by `EnsureDeleted`.

kopiczko:
I merge it now. I need it to unblock myself. We can discuss the log thing later.

PR #153, opened 2018-10-24
filelogger: make it idempotent

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4481.


pkg/internal/filelogger/filelogger.go
```diff
@@ -0,0 +1,127 @@
...
+			return microerror.Mask(err)
+		}
+
+		f.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("created log stream for pod %#q in namespace %#q", name, namespace))
+	}
+
+	{
+		f.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("starting logging job for pod %#q in namespace %#q", name, namespace))
+
+		go f.scan(ctx, logStream, logFilePath)
```
xh3b4sd: When will this goroutine actually be terminated?

kopiczko: When the `io.Copy` returns. Which is when the logged pod is deleted.

kopiczko: This was implemented like that so far.

xh3b4sd (approved):
Alright. Makes sense.

PR #155, opened 2018-10-25
release: setup logging when ensuring operator installed

kopiczko:

### DIFF
--- a/pkg/release/release.go
+++ b/pkg/release/release.go
@@ -279,6 +279,10 @@ func (r *Release) Install(ctx context.Context, name string, version Version, val
 	return nil
 }
 
+// TODO Remove once we are done with migration to EnsureInstalled.
+//
+//	Issue https://github.com/giantswarm/giantswarm/issues/4355.
+//
 func (r *Release) InstallOperator(ctx context.Context, name string, version Version, values string, crd *apiextensionsv1beta1.CustomResourceDefinition) error {
 	err := r.Install(ctx, name, version, values, r.condition.CRDExists(ctx, crd))
 	if err != nil {
--- a/pkg/release/release.go
+++ b/pkg/release/release.go
@@ -3,6 +3,7 @@ package release
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
@@ -173,10 +174,14 @@ func (r *Release) EnsureDeleted(ctx context.Context, name string, conditions ...
 }
 
 // EnsureInstalled makes sure the release is installed and all conditions are
-// met.
+// met. If release name ends with "-operator" suffix it also selects
+// a "app=${name}" pod and streams it logs to the ./logs directory.
 //
 // NOTE: It does not update the release if it already exists.
 func (r *Release) EnsureInstalled(ctx context.Context, name string, version Version, values string, conditions ...func() error) error {
+	var err error
+	isOperator := strings.HasSuffix(name, "-operator")
+
 	{
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating release %#q", name))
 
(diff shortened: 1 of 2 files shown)


pkg/release/release.go
```diff
@@ -184,10 +185,14 @@ func (r *Release) EnsureDeleted(ctx context.Context, name string, conditions ...
...
 
 // EnsureInstalled makes sure the release is installed and all conditions are
-// met.
+// met. If release name ends with "-operator" suffix it also selects
+// a "app=${name}" pod and streams it logs to the ./logs directory.
 //
 // NOTE: It does not update the release if it already exists.
 func (r *Release) EnsureInstalled(ctx context.Context, name string, version Version, values string, conditions ...func() error) error {
+	var err error
+	isOperator := strings.HasSuffix(name, "-operator")
```
kopiczko: Not sure if this is the best approach on earth. But previous one was also based on heavy assumptions.

MarcelMue: What is the benefit of doing this insteal of `InstallOperator` and using `EnsureInstalled` inside?

MarcelMue: To me this does not seem intuitive, when I call `InstallOperator` I expect something different to happen than just calling `Install`. Here there is no indication that something different is happening behind the scenes.

kopiczko: I want to have idempotent version of installing an operator. Also having consistency is nice. What's the alternative? `EnsureOperatorInstalled`? 

To make the logging nice I would write exactly like that:

```
func (r *Release) EnsureInstalled(ctx context.Context, name string, version Version, values string, conditions ...func() error) error {
	err := r.ensureInstalled(ctx, name, version, values, false, conditions...)
	if err != nil {
		return microerror.Mask(err)
	}
	
	return nil
}

func (r *Release) EnsureOperatorInstalled(ctx context.Context, name string, version Version, values string, conditions ...func() error) error {
	err := r.ensureInstalled(ctx, name, version, values, true, conditions...)
	if err != nil {
		return microerror.Mask(err)
	}
	
	return nil
}

func (r *Release) ensureInstalled(ctx context.Context, name string, version Version, values string, isOperator bool, conditions ...func() error) error {
}
```

Is that better?

MarcelMue: I don't get why you want to mix both things in `ensureInstalled`. `EnsureOperatorInstalled` adds additional functionality around the base functionality of `EnsureInstalled`. 

May solution would be to simply replace this line here with `EnsureInstalled`:
https://github.com/giantswarm/e2e-harness/blob/9f97abe68c8fb2ce7a50886098d4d4d24d187583/pkg/release/release.go#L283

kopiczko: I'd like to setup logging before we start waiting for CRD. Otherwise we can miss some error.

kopiczko: I mean error during CRD creation in the operator. That was a flaw of previous implementation.

MarcelMue: Okay I see your point. I still would not want two functions doing the same thing in different ways (`InstallOperator` and `EnsureInstalled`). So if that is a problem then let's fix it for everything.

pkg/release/release.go
```diff
@@ -201,6 +206,38 @@ func (r *Release) EnsureInstalled(ctx context.Context, name string, version Vers
...
+		err := r.waitForConditions(ctx, c)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensured operator pod exists for release %#q", name))
+	}
+
+	var operatorPodName string
+	var operatorPodNamespace string = r.namespace
```
MarcelMue: I think you might need to same fallback as `InstallOperator` to the GS namespace. I am not sure if the namespacing is right everywhere currently.

pkg/release/release.go
kopiczko: I'd like to get rid of `InstallOperator` completely.
I'd also like to remove `conditions` form `Install` and `Upadate` (where they aren't used btw) and make those methods lean. We should use only `Ensure*` implementations unless we have very special cases. In such cases I think we can use conditions explicitly.

pkg/release/release.go
kopiczko: Yes. I thought it's a good opportunity to test if it is right.

pkg/release/release.go
kopiczko: > I'd like to get rid of InstallOperator completely.

Maybe it's a good idea to do this in this PR?

MarcelMue: I agree, let's remove `InstallOperator` in this PR then and vendor this into our major operators. If everything goes well, then we can merge here.

pkg/release/release.go
```diff
@@ -242,6 +279,10 @@ func (r *Release) Install(ctx context.Context, name string, version Version, val
 	return nil
 }
 
+// TODO Remove once we are done with migration to EnsureInstalled.
+//
+//	Issue https://github.com/giantswarm/giantswarm/issues/4355.
+//
```
kopiczko: @MarcelMue I added TODO instead. This is the same as approach in `framework/resource` in general.

pkg/release/release.go
kopiczko: https://github.com/giantswarm/e2e-harness/pull/155#discussion_r228425284

kopiczko:
@MarcelMue PTALA

MarcelMue (approved):
Let's go with this, the follow up is mentioned in the issue so this is fine.

PR #156, opened 2018-10-25
release: fix setting up FileLogger

kopiczko:
https://github.com/giantswarm/giantswarm/issues/4481

FileLogger can't be set up externally because it's in internal package.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN chown -R e2e-harness:e2e-harness ${WORKDIR} ${HOME}
 RUN apk -Uuv add --update --no-cache \
       bash=4.4.19-r1 \
       build-base=0.5-r1 \
-      git=2.18.0-r0 \
+      git \
       jq=1.6_rc1-r1 \
       less=530-r0 \
       libffi-dev=3.2.1-r4 \
--- a/pkg/release/release.go
+++ b/pkg/release/release.go
@@ -107,8 +107,8 @@ func New(config Config) (*Release, error) {
 	var fileLogger *filelogger.FileLogger
 	{
 		c := filelogger.Config{
-			K8sClient: k8sClient,
-			Logger:    c.Logger,
+			K8sClient: config.K8sClient,
+			Logger:    config.Logger,
 		}
 
 		fileLogger, err = filelogger.New(c)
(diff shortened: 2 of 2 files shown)


pkg/release/release.go
```diff
@@ -107,15 +104,29 @@ func New(config Config) (*Release, error) {
...
+		c := filelogger.Config{
+			K8sClient: k8sClient,
+			Logger:    c.Logger,
+		}
+
+		fileLogger, err = filelogger.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
```
kopiczko: This is cool for me that using internal package enforced simplified external setup.
I didn't even know that filelogger is imported in dependent projects.

pkg/release/release.go
```diff
@@ -38,13 +37,14 @@ type Config struct {
...
 	apprClient *apprclient.Client
-	fileLogger *filelogger.FileLogger
 	helmClient *helmclient.Client
 	k8sClient  kubernetes.Interface
 	logger     micrologger.Logger
 
-	condition *conditionSet
 	namespace string
+
+	condition  *conditionSet
```
MarcelMue: Why do we need the third block now?

Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN chown -R e2e-harness:e2e-harness ${WORKDIR} ${HOME}
 RUN apk -Uuv add --update --no-cache \
       bash=4.4.19-r1 \
       build-base=0.5-r1 \
-      git=2.18.0-r0 \
+      git \
```
MarcelMue: I don't get this, why do we have these hard coded versions here anyway? Could we just leave them out completely?

pkg/release/release.go
kopiczko: Internals.

MarcelMue: The order should be `deps`, `internals`, `settings` then according to fmt https://github.com/giantswarm/fmt/blob/master/go/struct_initialization.md

kopiczko: It makes it harder to see what is passed from the config. But fine...

MarcelMue (approved):
LGTM

Dockerfile
xh3b4sd: Well, why did it change when it is unrelated? Why not just let it as it is?

kopiczko: I had failing build https://gigantic.slack.com/archives/C3C7ZQXC1/p1540449144000100

PR #160, opened 2018-11-06
release: allow installing multiple releases of the same chart

kopiczko:


xh3b4sd (approved):
I think this works.

kopiczko:
I don't understand why `e2eRemoteTest` fails out of the blue. We have the same situation here https://github.com/giantswarm/e2e-harness/pull/161#issuecomment-436314702.

I don't think those changes are related. I'm merging to unblock you.

PR #161, opened 2018-11-06
framework: improve latest/previous version inference (previous wip/current)

kopiczko:
After chat with @xh3b4sd we decided our current nomenclature for wip and
current version doesn't make much sense. Especially in the advent of
versioning of operators. Also what we want to do is basically always
test two latest versions and upgrades from the previous to the latest.
It would be cool to do more upgrade checks but it takes resources and
this thing here seems to be reasonable choice.


xh3b4sd (approved):
I think this is more right than what we have right now. This is still far from perfect though. In case we introduce patches within the chain of ordered versions we do not test them anyway. So SIG releng should come up with something here, also with regards of how all this works when we split up operators and run multiple ones with one version each.

kopiczko:
I don't understand why `e2eRemoteTest` fails here. This is the same in https://github.com/giantswarm/e2e-harness/pull/160. I smell some `shipyard` magic.

pkg/framework/version_bundle.go
```diff
@@ -147,12 +147,22 @@ func extractAuthorities(content, vType string) ([]versionbundle.Authority, error
...
 	sortedReleases := versionbundle.SortIndexReleasesByVersion(indexReleases)
 	sort.Sort(sort.Reverse(sortedReleases))
-	for _, ir := range sortedReleases {
-		if vType == "wip" && !ir.Active || vType == "current" && ir.Active {
-			return ir.Authorities, nil
-		}
+	switch vType {
+	case "latest", "wip":
+		return sortedReleases[0].Authorities, nil
+	case "previous", "current":
```
kopiczko: I thought about replacing `previous` with `stable`. It makes more sense when we have versioning of operators in place. Also it has the same length as `latest` which will look nice in AWS console :D 

I'm not big on that though. WDYT?

xh3b4sd: `stable` suggests some version semantic like `wip` and `cur`, which are both not correct. I feel the same alongside with `stable`. This is actually about numeric order. We are interested in the transitioning from one version to another. Regardless their semantic classification. We want to ensure that we are able to move on. This is about order. `latest` and `previous` makes sense here.

kopiczko: Now yes. But probably not when we have versioning of operators in place. But then we probably have to change this code anyway.

kopiczko:
Going despite `e2eRemoteTest` failing. This is unrelated. We should test actual changes in aws-operator.

PR #167, opened 2018-11-22
cmd, pkg: use dockerized go 1.11.1 for compiling binaries

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4753.

Before that e2e-harness was using OS go binary. On circleci this means
go 1.9. This means client-go@1.12.2 won't compile. It could have also
other side effects.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 e2eTest: &e2eTest
   machine: true
-  working_directory: /home/circleci/.go_workspace/src/github.com/giantswarm/operatorkit
+  working_directory: /home/circleci/.go_workspace/src/github.com/giantswarm/e2e-harness
   steps:
     - checkout
     - attach_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,12 +7,7 @@ e2eTest: &e2eTest
         at: .
     - run: ./e2e-harness localkube
     - run: ./e2e-harness setup --remote=false
-    - run:
-        command: |
-          echo "debug: TEST_DIR=$TEST_DIR GOPATH=$GOPATH"
-          ./e2e-harness test --test-dir=${TEST_DIR}
-        environment:
-          GOPATH: /home/circleci/.go_workspace
+    - run: ./e2e-harness test --test-dir=${TEST_DIR}
     - run:
         command: ./e2e-harness teardown
         when: always
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,11 @@ e2eTest: &e2eTest
     - run: ./e2e-harness localkube
     - run: ./e2e-harness setup --remote=false
     - run:
+        command: echo "debug: TEST_DIR=$TEST_DIR GOPATH=$GOPATH A=$A"
+        environment:
+          A: a_value
+          GOPATH: /home/circleci/.go_workspace
+    - run:
         command: ./e2e-harness test --test-dir=${TEST_DIR}
         environment:
           GOPATH: /home/circleci/.go_workspace
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,26 +1,40 @@
+e2eTest: &e2eTest
+  machine: true
+  working_directory: /home/circleci/.go_workspace/src/github.com/giantswarm/operatorkit
+  steps:
+    - checkout
+    - attach_workspace:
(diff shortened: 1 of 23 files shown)


pkg/tasks/tasks.go
```diff
@@ -1,23 +1,21 @@
...
 		}
 	}
 	return nil
 }
-
-func RunIgnoreError(tasks []Task) {
-	for _, task := range tasks {
-		task()
-	}
-}
```
kopiczko: Ninja. This wasn't used at all and looks wrong by name :D

cmd/localkube.go
```diff
@@ -46,5 +50,5 @@ func runLocalkube(cmd *cobra.Command, args []string) error {
 		l.SetUp,
 	}
 
-	return tasks.Run(bundle)
+	return tasks.Run(ctx, bundle)
```
kopiczko: `ctx` everywhere.

pkg/internal/docker/run.go
```diff
@@ -0,0 +1,48 @@
...
+	args = append(args, config.Image)
+
+	for _, arg := range config.Args {
+		args = append(args, arg)
+	}
+
+	err := exec.Exec(ctx, prog, args...)
+	if err != nil {
+		return microerror.Mask(err)
+	}
```
kopiczko: `Exec` exec defined below.

kopiczko:
Interesting:

```
Error: expected current working directory `/home/circleci/.go_workspace/src/github.com/giantswarm/e2e-harness` to be in GOPATH `/home/circleci/.go_workspace:/usr/local/go_workspace`: execution failed error
Usage
```

I didn't know you can have multiple directories in `GOPATH`. I'll deal with that.

tuommaki (approved):
Code LGTM. Good to go when tests are green. :heart:

.circleci/config.yml
```diff
@@ -1,71 +1,55 @@
+e2eTest: &e2eTest
+  machine: true
+  working_directory: /home/circleci/.go_workspace/src/github.com/giantswarm/operatorkit
```
xh3b4sd: Why `operatorkit`?

pkg/internal/golang/go.go
```diff
@@ -0,0 +1,108 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/giantswarm/e2e-harness/pkg/internal/docker"
+	"github.com/giantswarm/microerror"
+)
+
+const (
+	dockerImage = "quay.io/giantswarm/golang:1.11.1"
```
xh3b4sd: That should be configurable via flags for the commands of interest. Then we do not have to rebuild the project to update the version.

pkg/internal/golang/go.go
```diff
@@ -0,0 +1,108 @@
...
+	{
+		hostWdir, err = os.Getwd()
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	var hostGoPath string
+	{
+		envVar := "GOPATH"
```
xh3b4sd: That is already defined above.

pkg/project/project.go
```diff
@@ -150,7 +150,7 @@ func (p *Project) CommonTearDownSteps() error {
 	return nil
 }
 
-func (p *Project) Test() error {
+func (p *Project) Test(ctx context.Context) error {
```
xh3b4sd: It would make sense to do these changes independently to have nicer PRs and less to confuse. We could just have had one PR upfront to get the interfaces straight and then fix the other things.

.circleci/config.yml
kopiczko: Should be `e2e-harness` of course.

pkg/internal/golang/go.go
kopiczko: It's fine to rebuild the projects. Setting flag would be a pain. We'd have to update all circle configs instead of rebuilding project once.

pkg/project/project.go
kopiczko: Yes. You know how it is. You want to fix sth then everything else attacks you. I'll try to create other PR.

kopiczko: Done.

kopiczko:
All green now. Moved `ctx` stuff to another PR.
PTALA.

xh3b4sd (approved):

PR #171, opened 2018-11-23
tasks: add retry task

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4753.

This is to enable retry pulling `go` docker image.


pkg/tasks/spec.go
```diff
@@ -0,0 +1,8 @@
+package tasks
```
indyfree: Nice to move it here :)

pkg/tasks/retry_task.go
```diff
@@ -0,0 +1,44 @@
...
+	t := func(ctx context.Context) error {
+		o := func() error {
+			err := config.Underlying(ctx)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			return nil
+		}
+		b := backoff.NewExponential(backoff.ShortMaxWait, backoff.ShortMaxInterval)
```
indyfree: Any reason why you choose `backoff.ShortMaxWait, backoff.ShortMaxInterval` in particular? Does it make sense to make it configurable at the Task level?

kopiczko: No reason. That made sense for my particular use-case. If we need to configure that we can add that to config and default to the short.

indyfree (approved):
LGTM

tuommaki (approved):
LGTM

PR #179, opened 2018-12-11
framework: improve Guest.WaitForNodesUp logs

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4692

### DIFF
--- a/pkg/framework/guest.go
+++ b/pkg/framework/guest.go
@@ -1,6 +1,7 @@
 package framework
 
 import (
+	"context"
 	"fmt"
 	"time"
 
@@ -228,8 +229,8 @@ func (g *Guest) WaitForGuestReady() error {
 	return nil
 }
 
-func (g *Guest) WaitForNodesUp(numberOfNodes int) error {
-	g.logger.Log("level", "debug", "message", "waiting for k8s nodes to be up")
+func (g *Guest) WaitForNodesUp(ctx context.Context, numberOfNodes int) error {
+	g.logger.Log("level", "debug", "message", fmt.Sprintf("waiting for %d k8s nodes to be in %#q state", numberOfNodes, v1.NodeReady))
 
 	o := func() error {
 		nodes, err := g.k8sClient.CoreV1().Nodes().List(metav1.ListOptions{})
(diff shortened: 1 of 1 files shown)


pkg/framework/guest.go
```diff
@@ -238,16 +239,20 @@ func (g *Guest) WaitForNodesUp(numberOfNodes int) error {
...
 		if len(nodes.Items) != numberOfNodes {
-			return microerror.Maskf(waitError, "worker nodes are still not found")
+			return microerror.Maskf(waitError, "found %d k8s nodes but want %d", len(nodes.Items), numberOfNodes)
 		}
 
+		var notReadyCnt int
 		for _, n := range nodes.Items {
 			for _, c := range n.Status.Conditions {
 				if c.Type == v1.NodeReady && c.Status != v1.ConditionTrue {
-					return microerror.Maskf(waitError, "worker nodes are still not ready")
```
kopiczko: This log in particular was very misleading. There is possibility that we have more nodes than expected in the k8s api e.g. when we scale down in draining test.

rossf7 (approved):

pkg/framework/guest.go
```diff
@@ -238,16 +238,20 @@ func (g *Guest) WaitForNodesUp(numberOfNodes int) error {
 		}
 
 		if len(nodes.Items) != numberOfNodes {
-			return microerror.Maskf(waitError, "worker nodes are still not found")
+			return microerror.Maskf(waitError, "found %d k8s nodes but want %d", len(nodes.Items), numberOfNodes)
 		}
 
+		var notReadyCnt int
```
xh3b4sd: Is that name on purpose? I would go with `notReadyCount`.

xh3b4sd (approved):

pkg/framework/guest.go
xh3b4sd: I think we use something like `WaitForNodesDown` when scaling down. There is another helper for that IIRC.

pkg/framework/guest.go
kopiczko: It's on purpose. It's pretty common abbreviation AFAIK. If you want provide suggestions and I change it.

MarcelMue (approved):
LGTM

PR #184, opened 2019-01-14
fixes failing image build by adding backoff

xh3b4sd:
Fixes https://github.com/giantswarm/giantswarm/issues/4582.

### DIFF
--- a/pkg/minikube/minikube.go
+++ b/pkg/minikube/minikube.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"strings"
 
+	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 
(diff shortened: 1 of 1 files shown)


indyfree (approved):
Nice! Also some cleaning up and aligning on the way :+1:

xh3b4sd:
@giantswarm/sig-releng there was some error handling missing when doing some magic with the Docker env. See the code in `master`: https://github.com/giantswarm/e2e-harness/blob/4a94ba016067a6192f9885361c9ece6ee9a593c8/pkg/minikube/minikube.go#L37. I added error handling and the code actually appears to consistently fail all the time. So that was never useful. What to do with this? I would simply drop it, but am not sure what purpose it may once could have had.

pkg/minikube/minikube.go
```diff
@@ -33,60 +31,30 @@ func New(logger micrologger.Logger, builder builder.Builder, tag string) *Miniku
 // BuildImages is a Task that build the required images for both the main
 // project and the e2e containers using the minikube docker environment.
 func (m *Minikube) BuildImages(ctx context.Context) error {
-	m.logger.Log("level", "info", "message", "getting minikube docker environment")
-	env, err := m.getDockerEnv()
```
kopiczko: The error was unhandled but this was useful for local testing. I'm using that from time to time.
I'm not sure how it works with localkube, but I think just fine. I think the error should be just handled here.

xh3b4sd: Do you mean we should keep the code I removed? Because this does not work at all. CI fails then. I removed it and now it works. What shall I do?

kopiczko (approved):

pkg/minikube/minikube.go
kopiczko: For the record. We thought, we talked, no one is sure. Let's delete it and if needed implement something we can explain.

PR #188, opened 2019-02-05
Return actual error while waiting for k8s API

MarcelMue:


xh3b4sd (approved):
That makes sense to me.

pkg/framework/guest.go
```diff
@@ -193,7 +193,7 @@ func (g *Guest) WaitForAPIUp() error {
 	o := func() error {
 		_, err := g.k8sClient.CoreV1().Services("default").Get("kubernetes", metav1.GetOptions{})
 		if err != nil {
-			return microerror.Maskf(waitError, "k8s API is still down")
+			return microerror.Mask(err)
```
kopiczko: Below we do

```
g.logger.Log("level", "debug", "message", err.Error())
```

Can this be instead changed to:

```suggestion
			return microerror.Maskf(waitError, err.Error())
```

I think knowing this is a waitError is useful.

PR #192, opened 2019-04-09
Remove unused resource package

MarcelMue:
This got moved to `pkg/release` - we remove it here to get an better overview of what is actually used.


pkg/release/chart_info.go
```diff
@@ -22,6 +22,14 @@ func NewStableChartInfo(name string) ChartInfo {
 	}
 }
 
+func NewChannelChartInfo(name string, channel string) ChartInfo {
```
MarcelMue: Needed for e2etests as it uses channels.

xh3b4sd (approved):

kopiczko (approved):

rossf7 (approved):

sslavic:
Rebased on top of latest master.

tomahawk28:
`chart-operator` is using `resource` pkg still, and it has to move to `pkg/release`. 
@MarcelMue How did you check this? Any migration did you do before you do this PR, so I can also look into it?

MarcelMue:
@tomahawk28 I did check this by vendoring it but it was in April. This was merged by Stevo - My assumption would be that you can use the new functions in place of the old ones. Uncertain as I would need to take a deeper look to get back into it.

PR #193, opened 2019-04-11
release: add CRDNotFound condition

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5278.

- Add CRDNotFound condition.
- Rename SecretNotExist to SecretNotFound.
- Use waitError instead of stillExistsError.

PR #194, opened 2019-04-11
dockerfile: update OS deps

kopiczko:
Builds were failing.

See https://github.com/giantswarm/e2e-harness/pull/193#issuecomment-482020304.


kopiczko:
This fixes builds. Merging. Should at me if you have comments.

PR #196, opened 2019-04-12
release: add PodNotFound condition

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5278.

### DIFF
--- a/pkg/release/condition_set.go
+++ b/pkg/release/condition_set.go
@@ -78,7 +78,7 @@ func (c *conditionSet) CRDNotFound(ctx context.Context, crd *apiextensionsv1beta
 				return backoff.Permanent(microerror.Mask(err))
 			}
 
-			return microerror.Maskf(waitError, "CRD %#q still exists", crd.Name, crd.Namespace)
+			return microerror.Maskf(waitError, "CRD %#q still exists", crd.Name)
 		}
 		b := backoff.NewExponential(backoff.ShortMaxWait, backoff.ShortMaxInterval)
 		n := backoff.NewNotifier(c.logger, ctx)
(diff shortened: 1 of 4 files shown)


pkg/release/condition_set.go
```diff
@@ -78,7 +78,7 @@ func (c *conditionSet) CRDNotFound(ctx context.Context, crd *apiextensionsv1beta
 				return backoff.Permanent(microerror.Mask(err))
 			}
 
-			return microerror.Maskf(waitError, "CRD %#q in namespace %#q still exists", crd.Name, crd.Namespace)
+			return microerror.Maskf(waitError, "CRD %#q still exists", crd.Name, crd.Namespace)
```
MarcelMue: How does this look like now?

pkg/release/condition_set.go
```diff
@@ -120,6 +120,32 @@ func (c *conditionSet) PodExists(ctx context.Context, namespace, labelSelector s
...
+			if apierrors.IsNotFound(err) {
+				return nil
+			} else if err != nil {
+				return microerror.Mask(err)
+			}
+			if len(pods.Items) != 1 {
+				return microerror.Maskf(waitError, "expected 1 Pod but got %d", len(pods.Items))
+			}
+
+			return microerror.Maskf(waitError, "Pod for label selector %#q in namespace %#q still exists", labelSelector, namespace)
```
MarcelMue: Here you still use the namespace in the Maskf

pkg/release/condition_set.go
kopiczko: CRD is cluster scope. I saw it somewhere but I don't remember.

pkg/release/condition_set.go
kopiczko: Pod is namespaced :)

pkg/release/condition_set.go
MarcelMue: Then remove the `crd.Namespace` from here :D

kopiczko: Ah ok :)
```suggestion
			return microerror.Maskf(waitError, "CRD %#q still exists", crd.Name)
```

MarcelMue (approved):

PR #199, opened 2019-05-17
update vendor deps

xh3b4sd:


Gopkg.toml
```diff
@@ -91,8 +91,8 @@ required = [
   version = "kubernetes-1.13.2"
 
 [[constraint]]
-  branch = "master"
   name = "github.com/google/go-github"
+  version = "v25.0.4"
```
xh3b4sd: Only bumping the Github lib here as I need that in another scope of my stories.

PR #200, opened 2019-06-05
Fix minikube RBAC mode configuration instruction

sslavic:
Since minikube 26.0, kubeadm is the default bootstrapper for minikube, localkube has been deprecated.
README.md instruction how to start minikube with RBAC enabled was using localkube configuration style which is not compatible with kubeadm; it would cause minikube startup to hang.
This PR fixes the instruction how to start minikube with RBAC enabled to be compatible with kubeadm i.e. with recent minikube versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ for each operation mode:
 RBAC enabled before running e2e-harness:
 
 ```
-$ minikube start --extra-config=apiserver.Authorization.Mode=RBAC
+$ minikube start --extra-config=apiserver.authorization-mode=RBAC
 ```
 
 * Remote: as stated above e2e-harness uses [shipyard](https://github.com/giantswarm/shipyard)


MarcelMue (approved):
Fine with me and should work.
We are in the process of switching to KIND  (https://github.com/kubernetes-sigs/kind) so don't invest too much time into fixing this repo.

xh3b4sd:
More hand on our kind tooling would be appreciated. I think Ross is in the lead there right now.

rossf7 (approved):
If this works then let's do it. But yes IMO the sooner we complete the migration from `minikube` to `kind` the better.

@xh3b4sd @MarcelMue Yes I'd like to push `e2ectl` and `kind`. 

@sslavic There are a couple of small changes I could do with some help with. Is that OK?
 
I'll post in the main ticket https://github.com/giantswarm/giantswarm/issues/4606 in a bit.

sslavic:
Just for the context, I noticed this while debugging broken build https://github.com/giantswarm/chart-operator/pull/237

PR #203, opened 2019-06-27
drop vault teardown as it is not used anymore

xh3b4sd:


pkg/framework/helm.go
```diff
@@ -1,24 +0,0 @@
...
-
-import (
-	"log"
-	"os"
-	"os/exec"
-	"strings"
-)
-
-// HelmCmd executes a helm command.
-func HelmCmd(cmd string) error {
```
xh3b4sd: We finally also got rid of the `helm` "shell out".

pkg/framework/host.go
```diff
@@ -48,7 +47,6 @@ type Host struct {
 
 	clusterID       string
 	targetNamespace string
-	vaultToken      string
```
xh3b4sd: That was not used for anything anymore as well.

MarcelMue (approved):

PR #205, opened 2019-07-31
Update dependencies

sslavic:
Towards https://github.com/giantswarm/giantswarm/issues/4606

Among other things this updates helmclient dependency to benefit from changes done in https://github.com/giantswarm/helmclient/pull/112, timeout and selector improvements.

Current take for using KIND in e2e tests is to use e2ectl for cluster lifecycle, but still to use e2e-harness for running of the test itself. While iterating on this approach I've encountered same issue that e2e tests currently fail with, timeouts waiting for installed components (in my case tiller) to become ready. So I'm trying to use latest helmclient in e2e-harness first, and if issue doesn't go away even with these related latest improvements, will investigate deeper.


rossf7 (approved):
LGTM.

Good to get these updated but you can merge vendor PRs without approval. Just request a review and then merge.

Writing up https://github.com/giantswarm/giantswarm/issues/4919 as docs is top of my docs list and we need it for sig-releng onboarding.

PR #211, opened 2019-11-11
Add deprecation notice to README

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7584

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,10 @@
 
 # e2e-harness
 
+**Deprecated** We are moving away from using e2e-harness in our e2e tests.
+We now provision KinD clusters instead of using minikube. This is done by [e2ectl](https://github.com/giantswarm/e2ectl).
+Shared setup logic for e2e tests is being moved to [e2esetup](https://github.com/giantswarm/e2esetup).
+
 Harness for custom kubernetes e2e testing.
 
 ## Getting Project


README.md
```diff
@@ -2,6 +2,10 @@
 
 # e2e-harness
 
+**Deprecated** We are moving away from using e2e-harness in our e2e tests.
+We now provision KinD clusters instead of using minikube. This is done by [e2ectl](https://github.com/giantswarm/e2ectl).
```
rossf7: @sslavic I know you're still working on KinD to use a Circle Orb. We can update this if things change.

sslavic: Yep, will remember to update it once reusable e2e-integration-test job is in architect-orb.

sslavic (approved):
LGTM

MarcelMue (approved):

README.md
rossf7: @sslavic Great, thx!

kopiczko: If we plan to use the orb long term what is the reason to not use kind binary directly?

rossf7: As long as we can still provide a config file to kind I'm fine with using the binary directly.

We use this currently for multi node clusters and we may need more settings in future. e.g extending the kubeadm config to enable feature gates. 

https://github.com/giantswarm/e2ectl/blob/451a0b4a6f6d020683bcc2999d79d85aa5e2ec1e/cmd/cluster/create/template.go

kopiczko: That doesn't seem to be a problem to me. Thanks for the heads up.

PR #213, opened 2019-11-27
Ensure giantswarm namespace exists

whites11:
See https://github.com/giantswarm/k8scloudconfig/pull/609

### DIFF
--- a/pkg/framework/guest.go
+++ b/pkg/framework/guest.go
@@ -25,7 +25,7 @@ const (
 )
 
 var (
-	requiredNamespaces = []string{"giantswarm"}
+	namespaces = []string{"giantswarm"}
 )
 
 type GuestConfig struct {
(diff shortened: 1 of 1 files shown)


pkg/framework/guest.go
```diff
@@ -264,3 +273,20 @@ func (g *Guest) WaitForNodesReady(ctx context.Context, expectedNodes int) error
 	g.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("waited for %d k8s nodes to be in %#q state", expectedNodes, v1.NodeReady))
 	return nil
 }
+
+func (g *Guest) EnsureNamespaces(ctx context.Context, requiredNamespaces []string) error {
```
omegas27: I would probably name that EnsureNamespacesExists to be more clear on what this method actually does.
Also, the requirednamespaces could be renamed to namespaces imo.
WDYT?

Also, do you have a code review guide I could check?

whites11: Thanks for the review. I'm still working on it (build fails on things I didn't work on).

whites11: > Also, do you have a code review guide I could check?

No sorry, don't know if it exists, maybe try asking on #ask-me-anything

whites11: Function and variable renamed

PR #217, opened 2020-03-03
Use micrologger v0.1.0

fiunchinho:
We need to pin the micrologger dependency to the v0.1.0 version which is the one using dep instead of newer versions that use go modules.


fiunchinho:
It seems the `alpine` packages that we were using are not available anymore so I need to update their versions.

rossf7 (approved):
> It seems the alpine packages that we were using are not available anymore

Good to update. e2e-harness is deprecated but we need to maintain it a bit longer.

